PR #3, opened 2016-07-21
Remove needless call-outs

ajs6f:


acoburn:
Can we close this?

ajs6f:
Yeah, this stuff is long gone.

PR #6, opened 2016-11-23
Add messaging specification

acoburn:


index.html
```diff
@@ -471,7 +471,148 @@ <h2 class='informative'>Implementation Patterns</h2>
 
       </section>
     </section>
-	
+
+    <section id="messaging-spi">
+      <h2>Messaging SPI</h2>
+      <p>Non-normative: the messaging SPI defines when messages are emitted by a Fedora implementation, the minimal set of data contained in those messages and how those data are serialized. These messages allow clients to react to repository changes
```
ajs6f: Maybe break this out into a `section class="informative"` instead of using the (only-human-readable) phrase "non-normative"?

ajs6f:
It's hard for me to tell because of the way git diffs it, but am I right in thinking that this is going to be a top-level section parallel to the other chapters (it certainly should be)?

acoburn:
Yes it is. I have been just viewing the index.html file in a web browser, which makes it easier to see that type of formatting.

PR #16, opened 2017-01-17
Clarify that messaging can be synchronous or asynchronous

ajs6f:
Clarify that messaging can be synchronous or asynchronous


ajs6f:
@acoburn Does this make sense to you? If so please merge.

acoburn (approved):
This looks good to me.

PR #21, opened 2017-02-01
Fix typos

zimeon:
I think these are simple typos, not anything controversial:

[Section 1.2](http://fcrepo.github.io/fcrepo-specification/#httpPATCH) (and two other places) uses `LDPRS` instead of [`LDP-RS`](https://www.w3.org/TR/ldp/#dfn-linked-data-platform-rdf-source)

[Section 3.3.1](http://fcrepo.github.io/fcrepo-specification/#202-digest) "If the compared digests do no match" -> "do not match"

[Section 5.2 and 5.3](http://fcrepo.github.io/fcrepo-specification/#begin): /The effect of the request, it has one,/The effect of the request, if it has one,/ -- 4 times

### DIFF
--- a/index.html
+++ b/index.html
@@ -38,7 +38,7 @@
                 {   name:       "Adam Soroka",
                     company:    "University of Virginia",
                     companyURL: "https://www.virginia.edu/"},
-								{   name:       "Joshua Westgard",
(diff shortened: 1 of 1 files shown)


acoburn:
Thanks @zimeon!

PR #26, opened 2017-02-01
Fix RFC anchors to consistently use uppercase

zimeon:
Case used in RFC in refs is inconsistent: RFC6690, RFC7089, rfc2017, rfc6892 -- if nothing else, the inconsistent case makes the sort order in the references odd. Have changed to consistently use uppercase.


ajs6f:
Thanks @zimeon !

PR #30, opened 2017-02-01
Change "Fedora Resource" to LDPR

zimeon:
[Section 6.2](http://fcrepo.github.io/fcrepo-specification/#message-data) introduces the term "Fedora Resource" (used twice only) where I think it should be just `LDPR`. If something else is intended then it will instead need definition.

PR also links LDPR term to the definition in 3 other places.

### DIFF
--- a/index.html
+++ b/index.html
@@ -735,8 +735,8 @@
           Repository start, stop and configuration change operations MAY cause a message to be emitted.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


ajs6f:
This is messaging stuff, so I'm leaving it to @acoburn .

acoburn:
👍

PR #31, opened 2017-02-01
Fix 202-digest-link to digest-link param

zimeon:
[Section 3.3 Persistence Fixity preamble](http://fcrepo.github.io/fcrepo-specification/#persistence-fixity) mentions a  `202-digest-link` token value (only occurrence of that string). Seems that the intention is to have the new `202-digest` token, and then the `digest-link` parameter.

(Perhaps was a left over from an earlier iteration that had two tokens, no parameter?)

### DIFF
--- a/index.html
+++ b/index.html
@@ -563,7 +563,7 @@
       <section id="persistence-fixity">
         <h2>Persistence Fixity</h2>
         <section id="persistence-fixity-preamble" class="informative">
-          <p>In the minimal case, a client that has an original digest value can verify persistence fixity by downloading the LDP-NR and calculating the checksum. To avoid the necessity of transferring a potentially large binary, this specification describes expectation tokens for the <code>Expect</code> header described in [[!RFC2616]]: 202-digest and 202-digest-link</p>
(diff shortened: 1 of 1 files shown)


ajs6f:
I believe this is correct, but I would like @barmintor (whose language that is) to confirm before I merge this. I'd also like to take the `202-` out of that. It doesn't seem to add any info or distinguish anything, but that's a separate thing.

barmintor:
The `202-` is intended to follow the pattern of `100-continue`, which is the most popularly implemented Expect value. We did collapse both scenarios into `202-digest`, so the preamble departure is an oversight on my part. :+1:

barmintor:
That is, the `202-` contains a status code hint for operations that should be considered successful.

ajs6f:
Okay, I'll leave that bit alone. But this seems to have some kind of merge conflict? (How did we get a merge conflict with only one file!)

acoburn:
The merge conflict is from my recent PR changing the relevant RFC.

IMO, if we add line breaks we can avoid these sorts of conflicts.

acoburn:
@ajs6f github seems to have a new feature where one can resolve conflicts directly through their webUI, which is what I just did.

PR #34, opened 2017-02-02
Messaging: Remove SPI, add note that interface will be implementation specific

### LINKED ISSUE #29: Messaging SPI - support required? and clarification
The start of [Section 6: Messaging SPI](http://fcrepo.github.io/fcrepo-specification/#messaging-spi) says "This section is non-normative." -- I assume this means just the intro paragraph and not the whole of section 6? If so then I'd move that paragraph into a 6.1 (like [5.1](http://fcrepo.github.io/fcrepo-specification/#atomic-operations-introduction) with that standard "This section is non-normative." line. That will make it clear that the SPI is still a required part of the overall specification.

The term "SPI" was new to me on reading this spec, I see from a little searching (e.g. http://stackoverflow.com/questions/2954372/difference-between-spi-and-api & https://en.wikipedia.org/wiki/Service_provider_interface ) that is comes from the Java world which perhaps explains why I didn't know it. I think it would be worth pointing out explicitly the intention here is to specify an pattern and a serialization, but that the actual interface will be implementation specific.

I'd be happy to propose a change if my understanding is correct.

zimeon:
In anticipation that there might be agreement around current comments so that this fixes #29

### DIFF
--- a/index.html
+++ b/index.html
@@ -703,15 +703,17 @@
     </section>
 
     <section id="messaging-spi">
-      <h2>Messaging SPI</h2>
+      <h2>Messaging</h2>
       <section class="informative">
+	<h3>Introduction</h3>
         <p>
-          Non-normative: the messaging SPI defines when messages are emitted by a Fedora
+          This section defines when messages are emitted by a Fedora
           implementation, the minimal set of data contained in those messages and how
           those data are serialized. These messages may occur synchronously or asynchronously
 	  with the API operations that cause them to be emitted. They are typically used to
 	  support external integrations. The structure of these messages draws upon the
 	  existing [[activitystreams-core]] and [[ldn]] specifications.
+	  The messaging interface is not defined here and will be implementation specific.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -702,7 +702,7 @@
       </section>
     </section>
 
-    <section id="messaging-spi">
+    <section id="messaging">
       <h2>Messaging</h2>
       <section class="informative">
 	<h3>Introduction</h3>
@@ -713,7 +713,8 @@
 	  with the API operations that cause them to be emitted. They are typically used to
 	  support external integrations. The structure of these messages draws upon the
 	  existing [[activitystreams-core]] and [[ldn]] specifications.
-	  The messaging interface is not defined here and will be implementation specific.
+	  An implementation is free to choose from any messaging technology so long as
+	  the messages conform to what is described in the following sections.
         </p>
       </section>
 


index.html
```diff
@@ -703,15 +703,17 @@ <h2 class='informative'>Implementation Patterns</h2>
...
+	<h3>Introduction</h3>
         <p>
-          Non-normative: the messaging SPI defines when messages are emitted by a Fedora
+          This section defines when messages are emitted by a Fedora
           implementation, the minimal set of data contained in those messages and how
           those data are serialized. These messages may occur synchronously or asynchronously
 	  with the API operations that cause them to be emitted. They are typically used to
 	  support external integrations. The structure of these messages draws upon the
 	  existing [[activitystreams-core]] and [[ldn]] specifications.
+	  The messaging interface is not defined here and will be implementation specific.
```
acoburn: I do not agree with this sentence. The messaging interface _is_ defined here and it consists of the next three sections. It defines the required data, the serialization format and when events are emitted. What else needs to be defined? Yes, an implementation has some room for interpretation, but by saying that a message must conform to the Activity Streams 2.0 specification, we're being pretty clear.

index.html
```diff
@@ -703,15 +703,17 @@ <h2 class='informative'>Implementation Patterns</h2>
     </section>
 
     <section id="messaging-spi">
-      <h2>Messaging SPI</h2>
+      <h2>Messaging</h2>
       <section class="informative">
+	<h3>Introduction</h3>
         <p>
-          Non-normative: the messaging SPI defines when messages are emitted by a Fedora
+          This section defines when messages are emitted by a Fedora
```
acoburn: By removing the "Non-normative" text, a person reading this wouldn't know that this is non-normative language. If this is, indeed, non-normative language (and it _is_ non-normative language), then some marker for a human reader should be there.

zimeon, line 710: This test isn't needed as by providing an \<h3> heading there is a _"This section is non-normative."_ line above the paragraph.

acoburn: Oh, right, I forgot that RESPEC adds that.

index.html
zimeon: OK, I struggled with wording here. The point for me is that the rest of the spec tells you how to interact (e.g. by doing HTTP GET etc.) as well as the model and data formats, but the messaging section does not include the how to interact part (what I called the interface). I think that is what @awoods was getting at by choice of SPI not API. How about "The choice of messaging framework will be implementation specific." or similar?

awoods: > "The choice of messaging framework will be implementation specific." 

This seems like a constructive addition.

acoburn: Right, we are definitely _not_ defining the messaging protocol or the broker technology. My thought is that by being silent on this point, we leave the door wide open for implementations, but if you feel that it would be clearer with some non-normative language that explicitly states that we're not defining the messaging tech here, that would be OK.

I don't think "messaging framework" is quite right, but I'd be OK with something like this:

> An implementation is free to choose from any messaging technology so long as the messages conform to what is described in the following sections.

index.html
```diff
@@ -703,15 +703,17 @@ <h2 class='informative'>Implementation Patterns</h2>
     </section>
 
     <section id="messaging-spi">
-      <h2>Messaging SPI</h2>
```
awoods: As you mentioned, @zimeon, the Messaging section is different than the others in that it details what a server implementation must provide, not how a user "interfaces" with the capability. 
The inclusion of "SPI" tries to highlight that difference. 
Is your opinion that the difference in nature of this section from the others need not be emphasized?

zimeon: I think (and there seemed to be support for the idea) that the use of "SPI" doesn't make the difference clear enough because SPI is not well understood (I just asked 3 developers and none of them gave me a difference between API and SPI that would have illuminated the distinction). I think the sentence added about choice of messaging technology is a better way to make that point.

(I just removed -spi from the section id in last commit too, hadn't noticed that before)

ajs6f:
I'm fine with removing the SPI term (which is definitely Java-centric), but just for the record, there is a difference and it's not that obscure: APIs are made of types, instances of which you _use_ (some other process gives them to you), SPIs are made of types, instances of which you _provide_ (you give them to other processes). The distinction isn't sharp. Most contracts have some of both kinds of behavior. But that's the basic difference. You _use_ an API, you _implement_ an SPI.

acoburn (approved):

PR #35, opened 2017-02-02
Consistent use of Link: rel=... to describe Link headers

### LINKED ISSUE #28: Link@rel - format and absolute URIs
Starting with [section 1.2](http://fcrepo.github.io/fcrepo-specification/#httpPATCH) there are `Link@rel="ldp:constrainedBy"` style mentions of Link headers. I think these should be written `Link: rel="http://www.w3.org/ns/ldp#constrainedBy"` etc. 

  * Maybe I'm missing something but I think the `Link@rel` style is just an xpath-like shorthand that originated when dealing with things like Atom, not the best way to describe HTTP link headers
  * There is no notion of URI prefixes in Link headers per [LDP](https://www.w3.org/TR/ldp/#ldpr-resource) or [RFC5988](https://tools.ietf.org/html/rfc5988#section-5.3) _"Note that extension relation types are REQUIRED to be absolute URIs in Link headers..."_. So I think any relation not in the [IANA Link Relations registry](http://www.iana.org/assignments/link-relations/link-relations.xhtml) should be an absolute URI.

zimeon:
Fixes #28

### DIFF
--- a/index.html
+++ b/index.html
@@ -123,7 +123,7 @@
           provide a corresponding response body containing information about which statements could
           not be persisted. ([[LDP]] 4.2.4.4 SHOULD becomes MUST). In that response the restrictions
           causing such a request to fail MUST be described in a resource indicated by a
-          <code>Link@rel="ldp:constrainedBy"</code> response header per [[LDP]] 4.2.1.6. A successful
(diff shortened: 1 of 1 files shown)


awoods:
Looks good :+1:

PR #42, opened 2017-02-06
Add Fedora Ontology

acoburn:


repository.owl
```diff
@@ -0,0 +1,29 @@
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
```
awoods: Would an `application/owl+xml` serialization make sense for better HTML/XSLT rendering?
https://www.w3.org/TR/owl2-xml-serialization/

acoburn: I have no interest in writing XML (I find it difficult to read and difficult to maintain), but if you'd like to convert this document, be my guest.

awoods: What is the mime-type of this current serialization?

acoburn: it is TURTLE: `text/turtle`

awoods: Would a `.ttl` extension make more sense?

acoburn: using `.ttl` would be fine. I've just changed it in a new commit.

awoods:
:+1:

awoods:
@ruebot : you should now be good going forward.

Issue #43, opened 2017-02-08
Revisit PUT for creating new resources


birkland:
The language in the current Fedora specification for using `PUT` to create resources from [1.4.3](http://fedora.info/spec/#httpPUTcreate) is:

> an implementation MUST accept HTTP PUT to create resources ([LDP] 4.2.4.6 MAY becomes MUST).

[Recent conversation on IRC](http://irclogs.fcrepo.org/2017-02-07.html) suggests:
* There may not have been consensus around MAY becoming MUST in relation to mandating support for `PUT` to create resources in the first place
* the spec should address the topic of containment in relation to creating resources via PUT.

Opinions on the relationship between PUT-for-creation and containment vary:
* Resources created by `PUT` ought to be un-contained by definition
* The specification should define conditions under which `PUT` can be to as an idempotent way to create resources that are contained in an existing container.
* The specification should remain silent, meaning `PUT` is implementation-specific with regards to containment 

Therefore, this issue is to decide whether the spec continues to say that support for `PUT` remains a _MUST_ in the specification, and figure out how it ought to address the topic of containment in relation to resources created by `PUT`

acoburn:
+1 on continuing to use MUST for create-on-PUT

As for containment, I am fine with the spec being silent about containment w/r/t create-on-PUT, but perhaps some non-normative language would be useful here. Maybe suggesting that implementations document how containment works in relation to these resources?

I, at least, want an implementation to have the following behavior:

  * A `PUT` to create a new resource at `/a/b` (given the existence of `/a` and provided that `/a is a LDPC`) would succeed and create a `ldp:contains` triple from `/a` to `/a/b`.
  * A `PUT` to create a new resource at `/a/b` (given the existence of `/a` but `/a` is not an LDPC) would fail and link to an appropriate `ldp:constrainedBy` resource.
  * A `PUT` to create a new resource at `/a/b/c/d` (given the existence of `/a` but not any other intermediate resources: `/a/b` and/or `/a/b/c`) would fail and link to an appropriate `ldp:constrainedBy` resource.

I wouldn't argue that _all_ implementations need to have that behavior (e.g. that's not how the community impl works), but don't want the behavior described above to be out of conformance of the spec.

ajs6f:
[...]

PR #44, opened 2017-02-09
LDP 5.2.4.1 SHOULD becomes MUST

ajs6f:


index.html
```diff
@@ -182,6 +182,10 @@
           be assigned the new type and the interaction model of the resource MUST be changed to the
           interaction model assigned to the new type by [[LDP]].
         </p>
+        <p>
+          A PUT request that creates a new resource MUST NOT, while so doing, change the containment
+          triples in any LDPC. ([[LDP]] 5.2.4.1 SHOULD becomes MUST)
```
acoburn: Shouldn't this be `[[!LDP]]` since it's a normative reference?

ajs6f: You know, I honestly don't understand what our pattern of use is with ! vs no-!. I am happy to use whatever you think is more appropriate.

acoburn: As I understand ReSpec (which is somewhat limited), `[[BLAH]]` means "informative" while `[[!BLAH]]` means normative. We're clearly inconsistent in the document, but in this case, I think we want `[[!LDP]]`.

ajs6f: +1 to turning over a new leaf.

acoburn (approved):
👍

awoods:
@barmintor : thoughts on this PR?

barmintor:
@awoods I continue to think PUT-to-create is a bad idea.

EDIT: sorry, I thought this was another issue.

ajs6f:
As I recall, @barmintor, your objection is that `create-on-PUT` seems hard to implement to you?

awoods:
Thanks, @barmintor .
There are at least three options:
1. Disallow PUT-to-create
2. PUT-to-create is LDP-contained resource
3. PUT-to-create is NOT LDP-contained resource

@barmintor : you are for (1)?
@ajs6f / @acoburn : you are for (3)?

ajs6f:
No one has suggested (1), @awoods. The closest anyone has come has been to suggest that we do nothing more than LDP and leave it optional.

awoods:
Generally speaking, in the absence of agreement, I would suggest we proceed with the most lenient option.

acoburn:
Personally, I like (1) or (2).

But given (3), in my impl I would ignore the guidance of:

> An implementation must accept HTTP PUT to create resources.([LDP] 4.2.4.6 may becomes must). 

as it would be superseded by:

> 4.2.1.6 LDP servers must publish any constraints on LDP clients’ ability to create or update LDPRs, by adding a Link header with an appropriate context URI, a link relation of http://www.w3.org/ns/ldp#constrainedBy, and a target URI identifying a set of constraints [RFC5988], to all responses to requests that fail due to violation of those constraints.

I agree that while `create-on-PUT` is a MUST and containment is a `MUST NOT`, the spec is much clearer, but I just wouldn't implement that. It's not that doing so is difficult to implement (it's not hard at all), it's that it generates a disconnected repository graph, and I have significant reservations about that.

barmintor:
@awoods no, I would definitely *not* advocate disallowing it. I don't think people who want it are wrong; I just think it ought not to be a requirement that your repo be disconnected. I wish there was a way to advertise it in OPTIONS, but alas.

@ajs6f I have a create behavior for POST that is well-defined. I support PUT, too, but it essentially treats the request as POST with less worrying about whether the "new" URI exists. Mandating that I make it less like POST is more code.

barmintor:
Pursuant to @acoburn, I think haggling about SHOULD and MUST is premature before the proposed specs are implemented, or before clients understand what is actually necessary. I think @acoburn and @ajs6f are right that PUT-to-create (can we make this shorter? PUST or something?) needs something like this to make sense.

awoods:
@acoburn : as a note, I too have concerns about a disconnected repository graph.
I suspect multi-tenancy can be addressed with authorization policies.

acoburn:
I think there are a lot of options for implementing multi-tenancy; I just think that `create-on-PUT` is a bad way of doing that.

ajs6f:
Disconnected content in the server is precisely the purpose and advantage of the requirement. You are confusing "server" and "repository". They are not the same. A repository is a connected region within a server.

@awoods I think you are being very naive to think that authorization (which works entirely on the basis of containment) can be used to disconnect regions.

I'm getting really tired of opening the same issue over and over and having the same discussions over and over.

acoburn:
I fully support the distinction @ajs6f makes about a Fedora server and a Fedora repository. That said, I do not think it is the role of a client to create repositories inside a server, which is why I don't like `create-on-PUT`

barmintor:
Well, I think it's a legitimate way to do it- it requires less endpoint magic. But it's also not what this ticket is about: **IF** you support PUT-to-create, I think you have to take a position on containment, and this one is at least consistent. But you could also support PUT-to-create, **only** in situations where it could be reconstructed as if it were POST + Slug. Is it worth disallowing the one approach? Is there a way to advertise the difference? Is this a documentation problem?

ajs6f:
It is "what this ticket is about" in the sense that it is a use case for #44. There are others (tiering, segregating API-X config from other content, etc.).

barmintor:
@ajs6f I just meant that a discussion of PUT-to-create broadly is a tangent to this ticket, which seems to be trying to make the spec more sensible when PUT-to-create is supported.

ajs6f:
@barmintor, this is a PR, not a ticket. Are you thinking of https://github.com/fcrepo/fcrepo-specification/issues/43 ?

ajs6f:
@barmintor So given the status quo, are you comfortable merging this and opening a separate discussion about `create-on-PUT` generally?

barmintor:
@ajs6f this clarification is fine, but it should not have the note taking recourse to LDP 5.2.4.1- I don't think that requirement can be read to mean "containment of the created resource" and make sense with the 409 requirement.

ajs6f:
How does "don't create a new containment relationship" possibly _not_ follow from "MUST NOT allow HTTP PUT to update a LDPC’s containment triples"? If you create the resource contained, you have definitely altered someone's containment triples.

barmintor:
@ajs6f here are some things I would observe: 1) the mention of create follows that spec and 2) the 409 requirement only makes sense if the resource in question already exists, or if the 409 is supposed to indicate that put-to-create is not supported (it mirrors 4.2.4.3-4)

I think the LDP spec's intention is that a LDPC's containment triples are added or deleted with the object resources, not by manipulating the LDPC via PATCH or PUT-to-replace. PUT-to-create is ambiguous, but I really don't think what we're saying here (while valuable) is strengthening that requirement.

ajs6f:
I don't agree that 409 implies a pre-existing resource, but honestly, I don't much care at this point. If we all agree that (sans the LDP reference) this PR clarifies the current wording in a useful way, I will remove the LDP reference and merge this damn thing.

no-reply:
:+1: to removing the LDP reference. I made this point at https://github.com/fcrepo/fcrepo-specification/issues/43#issuecomment-278873163, but I'll repeat it here: __5.2.4.1__ reflects requirements for PUT requests sent to Container resources, not to new resources. The text is otherwise fine with me.

PR #45, opened 2017-02-10
normalize reference types

acoburn:


ajs6f:
This seems reasonable to me, but I don't use ReSpec a lot. @no-reply, I know you have more experience with it. Does this seem like the right policy?

ruebot (approved):

no-reply:
Looks good. :+1:

no-reply:
I don't have commit rights on this repo, or I'd merge it. :)

acoburn:
@ruebot should be all set now. Thanks!

PR #47, opened 2017-02-10
clarify restrictions on LDP interaction model changes via PATCH

### LINKED ISSUE #20: Clarification of interaction model changes

acoburn:


index.html
```diff
@@ -142,6 +142,11 @@
           A successful PATCH request MUST respond with a 2xx status code; the specific code in the 2xx
           range MAY vary according to the response body or request state.
         </p>
+        <p>
+          When receiving a PATCH request that would change the LDP interaction model of a resource
```
awoods: The essence of this text looks good. I would suggest tying it into the text above (`If an otherwise valid HTTP PATCH request is received that attempts to add statements to a resource that a server disallows...`) by moving the new text to a subsection 1.2.1, and re-phrasing as:
```
The server MUST disallow a PATCH request that would change the LDP interaction model of a resource
to a type that is not a subtype of the current resource type, the request MUST be rejected
with a 409 Conflict response.
```

acoburn: @awoods your suggestion seems fine to me -- would you be willing to submit a PR against this PR with your suggested change? I'm not quite sure what you have in mind w/r/t subsection structure. (I'd also note that too much re-organization will result in merge conflicts with #45)

awoods:
:+1:

awoods:
@ruebot , yes please... if you agree with the update.

PR #48, opened 2017-02-10
remove superfluous statement about content body

acoburn:
This removes the text: "containing a corresponding explanatory message in the response body".

Based on LDP 4.2.1.6, such constraints should be defined in a document linked to by a ldp:constrainedBy Link header, and as such defining such constraints in the response body seems unnecessary (certainly unnecessary to call out in the spec IMO).


awoods:
Given the fact that clients are informed of the reason for the 409 with the ldp:constrainedBy header... :+1:  on this PR.

ruebot (approved):

PR #53, opened 2017-02-10
Unsupported Digest types; Resolves #27

### LINKED ISSUE #27: Unsupported Digest types
[Section 1.4.1](http://fcrepo.github.io/fcrepo-specification/#httpPUTLDPNR) says:

> A HTTP PUT request that includes a Digest header (as described in [RFC3230](https://tools.ietf.org/html/rfc3230#section-4.3.2)) for which the instance-digest in that header does not match the instance it describes, must be rejected with a 409 Conflict response.

  * What is the appropriate response if the API implementation does not support one or more of the digest-algorithms sent in the client request?
  * Is it allowed to specify multiple digests with different algorithms (RFC3230 permits this, text about suggests just one with "the")?

It seems unreasonable to insist that all Fedora API implementations implement all Digest types (even just those in the RFC, certainly not all of the expanding [IANA registry]( http://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml)). Should there be a rule about handling the case of an unsupported Digest type (e.g. give 400 Bad Request?). Simply ignoring unsupported digests (per RFC3230 "A recipient MAY ignore any or all of the instance-digests in a Digest header field.") seems to defeat the point of having it for upload unless one relies on out-of-band understanding of supported types. Should there be a discovery mechanism?

(Obviously one could place a restriction on digest types that _MUST_ be supported, but that seems not the style of this spec and is not very future proof or extensible.)

ruebot:
* Stronger wording indicating a 400 status code
* Use @zimeon's suggested wording to indicate that multiple algorithms can be provided

See: https://github.com/fcrepo/fcrepo-specification/issues/27

### DIFF
--- a/index.html
+++ b/index.html
@@ -197,8 +197,12 @@
           </p>
           <p>
             A HTTP PUT request that includes a <code>Digest</code> header (as described
-            in [[!RFC3230]]) for which the instance-digest in that header does not match
-            the instance it describes, MUST be rejected with a 409 Conflict response.
+            in [[!RFC3230]]) for which any instance-digest in that header does not match the instance
+            it describes, MUST be rejected with a 409 Conflict response.
+          </p>
+          <p>
+            A HTTP PUT/POST request that includes an unsupported <code>Digest</code> type (as described
+            in [[!RFC3230]]), SHOULD be rejected with a 400 Bad Request response.
           </p>
           <p>
             Implementations MUST support <code>Content-Type: message/external-body</code> extensions
--- a/index.html
+++ b/index.html
@@ -164,9 +164,13 @@
         </p>
         <section id="httpPOSTLDPNR">
           <p>
-            A POST request that would create a LDP-NR and includes a <code>Digest</code> header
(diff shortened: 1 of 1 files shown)


acoburn (approved):
👍

index.html
```diff
@@ -197,8 +197,12 @@
           </p>
           <p>
             A HTTP PUT request that includes a <code>Digest</code> header (as described
-            in [[!RFC3230]]) for which the instance-digest in that header does not match
-            the instance it describes, MUST be rejected with a 409 Conflict response.
+            in [[!RFC3230]]) for which any instance-digest in that header does not match the instance
+            it describes, MUST be rejected with a 409 Conflict response.
+          </p>
+          <p>
+            A HTTP PUT/POST request that includes an unsupported <code>Digest</code> type (as described
```
awoods: The essence of this update is good, but instead of merging `PUT/POST` under the `HTTP PUT 1.4.1` section, I would suggest adding a similar update to the parallel text under HTTP POST 1.3``.

zimeon (approved):
:+1:

PR #54, opened 2017-02-10
SHOULD support Content-Type...; Resolves #37

### LINKED ISSUE #37: Requiring support for `Content-Type: message/external-body`
In [section 1.3](http://fedora.info/spec/#h-httppost), HTTP POST:

> Implementations **MUST** support Content-Type: message/external-body extensions for request bodies for HTTP POST that would create LDP-NRs. This content-type requires a complete Content-Type header that includes the location of the external body, e.g Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\", as defined in [rfc2017].

I'm curious as to the motivation for calling out this specific content type?  I initially read it as: a Fedora implementation, upon receiving a POST with `Content-Type: message/external-body` would dereference the URL and ingest the content of that URL (as an LDP-NR?).  Later I understood that the use case was to support the idea of Fedora 3 external data streams.

I wonder if it would be better for this to be a **MAY**, instead of a **MUST**, or simply remove this paragraph all together?  It seems odd to me that the specification requires _all_ Fedora implementations to carry over a use case from Fedora 3.  Or, I may just not fully understand the motivation.

ruebot:
See: https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-279026101

### DIFF
--- a/index.html
+++ b/index.html
@@ -169,7 +169,7 @@
             match that of the new LDP-NR MUST be rejected with a 409 Conflict response.
           </p>
           <p>
-            Implementations MUST support <code>Content-Type: message/external-body</code> extensions for
+            Implementations SHOULD support <code>Content-Type: message/external-body</code> extensions for
             request bodies for HTTP <code>POST</code> that would create LDP-NRs. This content-type requires
             a complete <code>Content-Type</code> header that includes the location of the external body, e.g
             <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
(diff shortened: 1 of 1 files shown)


acoburn (approved):
👍

birkland:
I'm good with this!

acoburn:
@ruebot I don't know why this has merge conflicts now, but can you rebase this PR?

acoburn:
@ruebot somehow the grammar doesn't make sense anymore.

acoburn:
It looks like the new line is added to the beginning of the wrong paragraph. It should replace the first line of the following paragraph.

acoburn:
Thanks @ruebot, that last rebase was much better.

PR #55, opened 2017-02-15
Reword abstract; Resolves #24.

### LINKED ISSUE #24: What is "Fedora information architecture"?
The [Abstract](http://fcrepo.github.io/fcrepo-specification/#abstract)'s use of term "Fedora information architecture" is undefined/unlinked and doesn't appear elsewhere in the document. This ends up making the statement seem a bit circular. From a little searching I didn't come up with anything that looked like a definitive document (and did find old docs like https://arxiv.org/abs/cs/0501012 which presumably gives the wrong picture). I think it needs a reference or rewording.

ruebot:
See: #24 

Pull together what I see to be the resolution to #24. However, I could not tell whether or not we want a reference to [this](https://arxiv.org/abs/1312.1258) paper there as well.

@awoods @zimeon @ajs6f @acoburn

### DIFF
--- a/index.html
+++ b/index.html
@@ -83,8 +83,8 @@
     <section id='abstract' class='informative'>
       <p>
         This specification refines the semantics and interaction patterns of [[LDP]] in order
-        to better serve the specific needs of those interested in implementing or using the
-        Fedora information architecture. Additionally, this specification contains:
+        to better serve the specific needs of those interested in implementing repositories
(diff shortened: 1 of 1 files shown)


ajs6f (approved):
👍

acoburn (approved):
👍

zimeon:
Belated :+1:

awoods:
Better belated than benevered.

PR #58, opened 2017-02-19
remove Date requirement from message data

acoburn:


ajs6f:
Is it still possible after this to use the event stream to journal another repo against the emitter?

acoburn:
Was that ever truly possible in the first place? I'm not convinced that it was.

ajs6f:
Okay, but there is now a canonical way to try, which (for a lot of cases, will work). Do we have a new suggestion for the use case of journaling?

acoburn:
You can trust the timestamp of the message from the message broker, which is going to be as reliable as a timestamp asserted by Fedora.

ajs6f:
Okay, that makes some sense.

awoods:
> It is entirely possible for an implementation to add dates, but they should not be required.

Given that some implementations will support timestamps in messages, it is helpful to have a consistently defined approach for doing so. Instead of striking this section, would it be reasonable to replace the `SHOULD` with `MAY`?

acoburn:
I'm fine with `MAY`, but I still object to this text:

> If the operation in question occurred as part of an atomic operation, the timestamps SHOULD correspond to the time the individual operation was performed (and not to the time the atomic operation was completed).

awoods:
Agreed, the atomic operation text is questionable. I am fine with changing this PR to:
1. remove the atomic operation text and
2. change `SHOULD` to `MAY`

acoburn:
@awoods and also the current spec document says that timestamps `MUST` be part of the event serialization. `MUST` is absolutely too strong; `SHOULD` is (IMO) also too strong.

birkland:
If the timestamp ~`SHOULD`~ `MAY` correspond to the time the resource was added/modified/deleted, then a client cannot assume any particular semantics to the date field without implementation-specific guidance.  In that case, I would agree with @acoburn  that it does not make sense for the text to be any stronger than "timestamps ~`MUST`~ `MAY` be part of the event serialization".

index.html
```diff
@@ -928,31 +923,11 @@ <h2 class='informative'>Implementation Patterns</h2>
...
-        "@type": "xsd:dateTime"
-      }
-    }
-  ],
+  "@context": "https://www.w3.org/ns/activitystreams",
   "id": "urn:uuid:3c834a8f-5638-4412-aa4b-35ea80416a18",
   "type": "Create",
   "name": "Resource Creation",
-  "date": "2016-05-19T17:17:39Z",
-  "actor": [
```
birkland, line 944: What does this change to "actor" mean in the context of this PR?  Was the former representation a typo?

acoburn: I can separate that into a separate commit if you'd like -- the issue is that the example there isn't really a "minimal" example.

birkland: Did you mean a separate issue and PR?  My instinct is that there may be some helpful non-normative text around actors, but I haven't really thought it through.  In any case, I don't want that to hold up this PR, so either way is fine.

acoburn: @birkland in this context, actors are simply taken from https://www.w3.org/TR/activitystreams-core/#actors; as such, the "minimal example" should show a "minimal" representation of an actor.

I will leave that change in place here unless someone specifically wants it moved into a separate PR.

birkland: Sounds good to me

zimeon (approved):
I agree that given vagueness of semantics, `MAY` seems appropriate

index.html
```diff
@@ -878,14 +878,6 @@ <h2 class='informative'>Implementation Patterns</h2>
         </p>
         <ul>
           <li>Resource Identifier: the URI of the <a>LDPR</a> that was added, changed or removed.</li>
-          <li>
-            Timestamp: a timestamp marking the time of the event. The value of the timestamp
```
awoods, line 882: Based on this [comment](https://github.com/fcrepo/fcrepo-specification/pull/58#issuecomment-281091019), I was expecting this text to now look like:

>       Timestamp: a timestamp marking the time of the event. The value of the timestamp
>       SHOULD correspond to the time the resource was added/modified/deleted, but there is no
>       strict requirement about the precision of this value.

acoburn: @awoods: that entire section is under data that `MUST` be included. I like the new text much better.

PR #59, opened 2017-02-20
Use existing ActivityStreams vocabulary for date example

acoburn:
Given: 

> Wherever possible, data should be expressed using the [activitystreams-vocabulary].

The example message should be serialized with `as:updated` instead of `dc:date`


awoods:
:+1:

awoods:
@ruebot : hit it if you are good with it.

PR #60, opened 2017-02-20
RFC 2396 is obsolete; we should use RFC 3986 instead

acoburn:


ruebot (approved):

zimeon:
This seems good as it is, but given the merge of #61 (URI->IRI), shouldn't we instead link to [RFC 3987](https://www.ietf.org/rfc/rfc3987.txt) which defines `IRI-reference` in a similar fashion?

index.html
```diff
@@ -669,7 +669,7 @@ <h2 class='informative'>Implementation Patterns</h2>
           <p>
             <code>instance-digest</code> values are of the syntax described in [[!RFC3230]] for the
             <code>Digest</code> header. <code>URI-reference</code> values for the
-            <code>digest-link-param</code> are described in [[!RFC2396]].
+            <code>digest-link-param</code> are described in [[!RFC3986]].
```
awoods: With the [change from URI to IRI](https://github.com/fcrepo/fcrepo-specification/pull/61/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR671), RFC2396 and RFC3986 no longer seem to apply.

https://www.ietf.org/rfc/rfc3987.txt seems more appropriate.

acoburn:
This PR now links to RFC 3987. Thanks, @zimeon!

zimeon (approved):

PR #61, opened 2017-02-20
Use IRI instead of URI

acoburn:
The Fedora specification should support the use internationalized identifiers, and so references to URI should be changed to IRI.

I left the use of `URI-R` and `URI-M` unchanged, as that is the terminology that Memento uses, though I think we should also consider changing those to `IRI-R` and `IRI-M`, respectively.

### DIFF
--- a/index.html
+++ b/index.html
@@ -658,17 +658,17 @@
           <h3>Expect: 202-digest</h3>
           <p>
             The <code>202-digest</code> expectation indicates that the client is requesting that
-            the server verify an instance digest for the resource at the request URI.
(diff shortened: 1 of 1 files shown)


ajs6f:
+1 to the main point, -1 to using `IRI-M` and `IRI-R` on our own. I appreciate the motivation, but it's more important to make the links to Memento. Instead, let's send a "PR" to that community to help them move to the New Slang. I'm sure HvS and Co. would be happy to, they probably just need a reminder.

ruebot (approved):

PR #63, opened 2017-02-22
Change definition of Prefer header IRIs

### LINKED ISSUE #62: Consider renaming prefer headers

acoburn:


index.html
```diff
@@ -259,11 +259,12 @@
           </p>
           <ul>
             <li>
-              <code>http://fedora.info/definitions/repository#EmbedResources</code>:
-              Requires a server to include representations of any contained resources in the response.
+              <code>http://www.w3.org/ns/oa#PreferContainedDescriptions</code>:
```
awoods, line 262: I like the idea of reusing existing vocabularies, but am wary of the slight disconnect between Fedora's notion of embedding resources and Open Annotation's notion of descriptions of contained Annotations:
http://w3c.github.io/web-annotation/vocab/wd/#prefercontaineddescriptions

...particularly since the definition of `PreferContainedDescriptions` specifically calls out: `...client prefers to receive full descriptions of the Annotations from a container...`

acoburn: Let me put this another way: say you had a Fedora/LDP implementation that also supported the WA spec. How would `oa:PreferContainedDescriptions` be different than `fedora:PreferEmbeddedResources` (or whatever the IRI is called)? In my reading of the WA spec, the two Prefer IRIs would be (functionally) equivalent, which makes me very suspicious of the need to create a separate IRI managed by Fedora.

awoods: @acoburn : that is a pretty reasonable argument. I will not have time to dig more deeply into the Web Annotations spec at the moment, but would be interested to hear if @azaroth42 sees the functional equivalence of the proposed scenario.

azaroth42: The intention of PreferContainedDescriptions is that the representation delivered in response to the request should embed sufficient information about the contained resources to ensure that they are usable by a good number of applications. It doesn't need to be every single triple, but it can't be just the URI (PreferContainedIRIs). That the protocol talks about "annotations" is because of the context, rather than any formal definition -- which is why we went with ContainedDescriptions, not ContainedAnnotations which would have made this use problematic.

awoods: Thanks, @azaroth42 . That seems very much in line with @acoburn's suggestion.
:+1:

awoods:
@fcrepo/committers : any further comments here? If not, let's move it forward.

PR #65, opened 2017-02-28
add a section on conformance

### LINKED ISSUE #64: Define levels of API conformance?
Should the spec, or an auxiliary document, define levels of conformance to the API? The main use case in my mind is having a Read-Only level, that only supports search and access. There may be other use cases.

acoburn:


zimeon (approved):
Great addition! While I might imagine more discussion about the _MUST_'s and _MAY_'s, I think it is clearly a win to make it clear that conformance means implementing all sections.

ajs6f (approved):
Good stuff.

acoburn:
And in case it isn't obvious, ReSpec adds this text to the conformance section:

> As well as sections marked as non-normative, all authoring guidelines, diagrams, examples, and notes in this specification are non-normative. Everything else in this specification is normative.
> 
> The key words may, must, must not, should, and should not are to be interpreted as described in [RFC2119].

zimeon (approved):

ruebot (approved):

PR #71, opened 2017-03-16
Reinforce explicit relationship to LDP 1.0

### LINKED ISSUE #25: Relationship to LDP
The (non-normative) abstract says this specification refines LDP, but I think there should be a _normative_ statement about what that means. Is it that "The Fedora API specification follows the LDP specification except where described in this document"?

It might also make sense to make a clear statement about the relation to versions of the LDP specification. The reference is to LDP 1.0 so I assume the statement is that this specification builds on and refines LDP 1.0 (and will still do so when LDP puts out their 1.1).

Various terms/acronyms in this document are used without definition which I assume means they are as defined in LDP. Should there be an explicitly linked statement along the lines of "Terms from LDP ([2 Terminology](https://www.w3.org/TR/ldp/#terms)) are used in this document."?

Finally, and I know this one is awkward: This specification would be about a zillion times easier to read/follow if the LDP section numbers references linked directly to the appropriate LDP sections. I understand the desire to do references the W3C way, but it really hampers usability. A compromise might be to keep the "\[LPD\]" linked to the references, but then link the actual section numbers to the points in the LDP spec?

Rarian:
* Adds a clarifying normative statement about conforming to LDP 1.0
* Adds a Terminology section
* Links LDP and Memento terms and references to their origins

Closes #25

### DIFF
--- a/index.html
+++ b/index.html
@@ -110,29 +110,94 @@
 
     <section id="conformance">
       <p>
-        A conforming <b>Fedora server</b> is a conforming [[!LDP]] server that also follows
-        the rules defined by Fedora in <a href="#resource-management"></a>, <a href="#resource-versioning"></a>,
-        <a href="#binary-fixity"></a>, <a href="#resource-authorization"></a>, <a href="#atomic-operations"></a>,
-        and <a href="#messaging"></a>.
+        A conforming <b>Fedora server</b> is a conforming [[!LDP]] 1.0 server except where described in this document
+        that also follows the rules defined by Fedora in <a href="#resource-management"></a>,
+        <a href="#resource-versioning"></a>, <a href="#binary-fixity"></a>, <a href="#resource-authorization"></a>, 
+        <a href="#atomic-operations"></a>, and <a href="#messaging"></a>.
       </p>
     </section>
-
+    <section id="terminology">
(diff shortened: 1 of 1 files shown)


acoburn (approved):
This all looks good to me. Thanks @Rarian!

awoods:
It looks like the following `LDP-NR` terminology reference was missed:
https://github.com/fcrepo/fcrepo-specification/pull/71/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR712

...otherwise, this looks great.

Rarian:
There doesn't seem to be a line number in that link, but I found an unlinked LDP-NR in the fixity section and added it.

awoods:
@Rarian : I believe you found the `LDP-NR`. Thanks.

On a side note, in the course of code reviews, it is helpful to respond to comments with new commits versus "push --force". With force pushes, it requires a full re-review since the differences are not apparent.
Once the code review is completed and all comments addressed, the person who merges the pull-request will squash/merge the incremental commits.

Rarian:
Fair enough, it's been a while since I contributed to a Fedora project, and my current projects have had the force push mentality, I'll endeavour to make subsequent pushes fixups in the future.

PR #72, opened 2017-03-17
Add security and privacy sections

### LINKED ISSUE #70: Privacy/Security section?
ReSpec gives this warning:

> This specification doesn't appear to have any 'Privacy' or 'Security' considerations sections. Please consider adding one, see https://w3ctag.github.io/security-questionnaire/

Should such a section be added?

acoburn:


index.html
```diff
@@ -1002,6 +1002,39 @@ <h2 class='informative'>Implementation Patterns</h2>
...
+      <h2>Security Considerations</h2>
+      <p>
+        There are security considerations for any HTTP server exposed to the Internet.
+        As such, implementations SHOULD make use of the many security-related features
+        of HTTP, including use of the authentication framework defined in the Authentication
+        chapter of the HTTP/1.1 specification [[!RFC7235]].
+      </p>
+      <p>
+        Fedora implementations are subject to the same security considerations as HTTP/1.1
+        ([[!RFC7230]] and [[!RFC7231]]), [[!LDP]], Memento ([[!RFC7089]]) and [[!activitystreams-core]].
```
awoods: In a similar way that we provide section links to `LDP` spec references, is there any reason not to do the same here?
- https://tools.ietf.org/html/rfc7230#section-9
- https://tools.ietf.org/html/rfc7231#section-9
- https://tools.ietf.org/html/rfc7089#section-7
- https://www.w3.org/TR/activitystreams-core/#security-considerations

index.html
```diff
@@ -1002,6 +1002,39 @@ <h2 class='informative'>Implementation Patterns</h2>
...
+      <h2>Privacy Considerations</h2>
+      <p>
+        Fedora resources may contain personally identifiable information such as the identify of
+        the agents who create and modify resources. As such, this information could be analyzed
+        to generate profiles of the behavior of individuals or groups.
+      </p>
+      <p>
+        As such, Fedora implementations that produce or consume personally identifiable
+        information MUST clearly document the type and extent of personal information that
+        is collected and published along with the reasons for doing so.
```
ajs6f: How is an implementation supposed to do this?

acoburn: @ajs6f an implementation would do this via documentation. For example, an implementation's documentation might read:

> User data is stored in a resource's audit record. Administrators should take care when exposing such information to the public web.

That would suffice, in my opinion.

That said, the text here is intended simply as a starting point, and if you have different ideas about a "Privacy" section (including dropping the section entirely), I would welcome comment.

index.html
acoburn: @awoods that makes sense to me, I can update the PR in the next couple of days.

index.html
ajs6f: I'm not dead-set against it, but I don't think it has much value and I think it is _way_ outside the remit of this spec to be discussing these issues in any particular way.

acoburn: @ajs6f much of what is said in the privacy section is there because of what exists in the privacy section of the [Activity Streams specification](https://www.w3.org/TR/activitystreams-core/#privacy) (which the Fedora spec makes use of):

> Implementations that produce or consume Activity Streams 2.0 documents must take steps to openly and publicly document and communicate to all potential users: (a) the kinds of potentially sensitive personal information published, consumed or collected by the implementation, (b) the reasons for publishing, consuming and collecting that information, (c) the manner in which that information is being used, (d) the identity of any other party with whom that information is being shared, and (e) the reason the information is being shared with other parties.
> 
> Implementations that publish Activity Streams 2.0 documents should assume a default position of limiting both the kind and amount of sensitive personal information included in the document unless users have "opted in" to sharing additional detail.

Given that a complying Fedora implementation would be an activity stream producer, it would necessarily follow that that implementation `MUST` include some documentation about any privacy considerations.

Alternately, Fedora could ignore the Activity Streams `MUST` statement, perhaps with it's own softer statement (e.g. `MAY`), but I would advocate for conforming with Activity Streams.

index.html
```diff
@@ -1002,6 +1002,39 @@ <h2 class='informative'>Implementation Patterns</h2>
       </section>
     </section>
 
+    <section id="privacy">
+      <h2>Privacy Considerations</h2>
+      <p>
+        Fedora resources may contain personally identifiable information such as the identify of
```
zimeon: identify -> identity

index.html
acoburn: @awoods is [this](https://github.com/fcrepo/fcrepo-specification/pull/72/commits/f938fb8a9c947e4b8b05b11dea2b13401d719526) what you had in mind?

index.html
awoods: Are we blocked on this?

Rarian: I think it makes sense to have a privacy section.  

I'll try to ponder some alternate verbiage that could hopefully address the initial question "How does an implementation do this?"

zimeon:
I like the security section as a set of links to appropriate sections of the underlying specs.

I feel the prescriptiveness of the privacy section is a bit of an overreach but I suppose with Activity Streams having that we are perhaps stuck. On style grounds I suggest at minimum removing the "As such, " that starts the second paragraph -- it seems very odd at the start of a paragraph and also having two sentences in a row starting that way. As an alternative suggestion one might trim the privacy section a little with an explicit link to the AS requirements, perhaps:

> Fedora resources, including records of agents who create and modify resources, 
may contain potentially sensitive personal information. Fedora implementations
are subject to the requirements of [[!activitystreams-core]] 
<a href="https://www.w3.org/TR/activitystreams-core/#privacy">6. Privacy Considerations</a>.
As such, Fedora implementations that produce or consume personally identifiable
information _MUST_ clearly document the type and extent of personal information that
is collected and published, along with the reasons for doing so.

acoburn:
For the privacy section, how about we just refer to the AS 2.0 spec rather than recapitulating what is said there. I.e. like what we do for the security section.

zimeon:
@acoburn - I'm fine with an even shorter link-out too

ajs6f:
The shorter the better. A single sentence to cover both would be enough.

ruebot (approved):

zimeon (approved):

PR #73, opened 2017-03-17
add clarifying statements

acoburn:
I don't always remember which RFC is which, and this adds a little bit of clarification with regard to references to the Memento spec.


whikloj:
I like this, its simple but makes it much easier to read.

PR #84, opened 2017-04-06
Add clarification of HTTP methods used with persistence fixity

### LINKED ISSUE #32: Say `Expect: 202-digest` intended for use with `HEAD`?
Related to #31, would it make sense to add to [Section 3.3 Persistence Fixity preamble](http://fcrepo.github.io/fcrepo-specification/#persistence-fixity) a statement that the `Expect: 202-digest` is intended to be used with `HEAD`? Also may be used with `GET` (though kinda pointless since one gets the bits in that case)?

zimeon:
Perhaps fixes #32

### DIFF
--- a/index.html
+++ b/index.html
@@ -713,7 +713,7 @@
         <section id="persistence-fixity-preamble" class="informative">
           <h2>Extension of <code>Expect</code> header</h2>
           <p>
-            Persistence fixity is verified by comparison of stored fixity results with newly
+            Persistence fixity is verified by comparison of previous fixity results with newly
             computed fixity results using <code>HEAD</code> or <code>GET</code> requests to <a>LDP-NR</a>s.
             In the minimal case, a client that has an original digest value can verify persistence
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -713,6 +713,8 @@ <h2 class='informative'>Vary</h2>
         <section id="persistence-fixity-preamble" class="informative">
           <h2>Extension of <code>Expect</code> header</h2>
           <p>
+            Persistence fixity is verified by comparison of stored fixity results with newly
+            computed fixity results using <code>HEAD</code> or <code>GET</code> requests to <a>LDP-NR</a>s.
```
ajs6f: Doesn't `stored` imply a little more than we want? Maybe just `previous`?

zimeon: My thought was that is has to be stored somewhere, in the client or the server, but I could go with `previous` --- will change

ruebot (approved):

ajs6f:
It's fine with me.

Issue #83, opened 2017-04-06
Support for proxied content (related to `Content-Type: message/external-body`)


kefo:
This issue is to focus and capture the debate around supporting a use case for proxied content. Substantial conversation on this topic happened as part of this issue - beginning at https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-290537966 - after it had been closed.

With respect to proxying @escowles [noted](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-290721251):

> Proxying seems like it would potentially address situations where Fedora could access the content but the client could not (either because of auth, private network, etc.).

> Proxying would also make access-type=local-file make sense for some implementations, but not all [as noted].

@barmintor will likely follow with additional details and use case.

barmintor:
@kefo I think the latest changes to the draft spec would allow closing this issue, can you confirm?

awoods:
@kefo : I believe this issue has been resolved with: https://github.com/fcrepo/fcrepo-specification/pull/121
If you agree, would you please close this issue?

awoods:
Resolved with https://github.com/fcrepo/fcrepo-specification/pull/121

Issue #80, opened 2017-03-30
`Content-Type: message/external-body` description insufficient for conformance tests


kefo:
The description for `Content-Type: message/external-body` lacks the details to confidently write an API conformance test.  Quoting [3.4.1](http://fedora.info/spec/#httpPUTLDPNR):

> Implementations MUST support Content-Type: message/external-body extensions for request bodies for HTTP PUT to LDP-NRs. This content-type requires a complete Content-Type header that includes the location of the external body, e.g Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\", as defined in [RFC2017]. 

As written, the header is noted to be an example, "e.g. Content-Type:..."  

Is `access-type=URL` the only supported type (assuming a restriction)?  Is [`access-type=local-file`](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-290537966) supported?  See [rfc2046.](https://tools.ietf.org/html/rfc2046#page-33)

Resolving https://github.com/fcrepo/fcrepo-specification/issues/40 is also needed.

acoburn:
A HEAD request on the resource could supply the content type. It is worth noting that a Fedora impl SHOULD (not MUST) support this. (Note: as long as the language for this is for redirect and not proxy, I plan not to support this in the impl I'm working on.)

awoods:
@kefo : I believe this issue has been resolved with: https://github.com/fcrepo/fcrepo-specification/pull/121
If you agree, would you please close this issue?

awoods:
Resolved with https://github.com/fcrepo/fcrepo-specification/pull/121

Issue #41, opened 2017-02-03
Offer guidance on an analog to external datastreams


barmintor:
Using `Content-type: message/external-body; access-type=URL;` for referenced data requires the client to have unmediated access to anything described at the URL. This is generally not the case for content historically described as 'external', for which a repository acts as a proxy.

awoods:
Agreed, the `external-body` approach here is more like the historical 'redirect' datastream type.

ajs6f:
+1 to Fedora <4 "redirect" being the right comparison.

ruebot:
@barmintor trying to understand on what needs to be done to resolve this issue, do we need to update the language in:
 - 1.3 http://fcrepo.github.io/fcrepo-specification/#httpPOST
 - 1.4.1 http://fcrepo.github.io/fcrepo-specification/#httpPUTLDPNR

...or is there more to it?

awoods:
@barmintor ^^

barmintor:
It's not clear to me what analog we can or should offer in the spec. I created this ticket in response to a call on which it was suggested that `external-body` was a way to migrate external datastreams, which it is not.

barmintor:
I think I also mentioned that it was generally an impediment to migration at Columbia, but that's not a spec issue.

awoods:
It does not appear that there is anything to be done towards resolution of this issue. @barmintor, if you agree, would you be willing to close it?

kefo:
I think this can be closed.  The issue here - clarification that the `message/external-body` resource is a "redirect" - will be resolved once https://github.com/fcrepo/fcrepo-specification/issues/40 is resolved.  It's a duplicate.

awoods:
@barmintor : I believe this issue has been resolved with: https://github.com/fcrepo/fcrepo-specification/pull/121
If you agree, would you please close this issue?

Issue #40, opened 2017-02-03
Clarify specification of external-body resources


birkland:
Pulled out of [#37](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-277297703)
- [ ] Decide on whether to specify [303](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-277093698)? [307](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-277089334)? or [3xx](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-277089969)?
- [ ] We text that "spells out what happens on HEAD", i.e. also redirect with 3?? (as decided above)
- [ ] Add non-normative clarifying text per: [#37 comment](https://github.com/fcrepo/fcrepo-specification/issues/37#issuecomment-277079418)

awoods:
Based on the points above, I would suggest:
1. 303
2. Same as 1.
3. Yes, add non-normative text

awoods:
@birkland : in light of: https://github.com/fcrepo/fcrepo-specification/pull/121
I believe this issue can be closed as "overcome by events".
If you agree, could you please close this issue?

awoods:
Resolved with https://github.com/fcrepo/fcrepo-specification/pull/121

birkland:
@awoods Sorry, I missed this.  Yes, I don't think this is relevant any more.  That being said, the external content section is not entirely comprehensible to me; At some point (after OR) I plan on opening tickets so that it can be clarified; albeit in a different way then this ticket.

PR #89, opened 2017-04-09
Change external-body support to SHOULD

acoburn:


awoods:
It would seem that the related, in-progress discussions around `external-content` should be addressed prior to changing relevant elements of the spec.
#40 
#41 
#80 
#83

acoburn:
@awoods: I disagree. All of those tickets relate to the details of the interaction model _if_ a given resource uses the `external-body` resource type. This is about whether that type is _required_ in the first place.

acoburn:
@awoods I will also note that Trellis will _not_ be supporting external-body resources. I would assert that there are better ways to deal with "external content", and _requiring_ a pattern that is arguably of dubious quality is a bad thing to have in the spec.

awoods:
@acoburn : Fair enough. Has there been discussion in support of the recommendation to remove the requirement of "external content" from the spec?

acoburn:
My claim is that the _requirement_ is ambiguous in the text: POST has _should_ and PUT has _must_. That is wrong. This PR aligns the language to the less stringent level; if there should be discussion on the matter, then please open an issue.

Issue #90, opened 2017-04-09
Does Persistence Fixity belong in the specification?


acoburn:

acoburn:
Before including a notion of persistence-fixity in the specification, I think it will be necessary to answer a few questions and provide a solid justification.

First, in what fundamental ways does persistence-fixity differ from transport-fixity? And what is it about persistence-fixity that cannot be handled by the existing transport-fixity mechanism (e.g. `Want-Digest`)?

Second, Fedora is concerned about resource durability, but what does that mean? What if an implementation was entirely in-memory? Would that mean that the implementation was, by definition, not a Fedora? What if that in-memory Fedora is distributed across lots and lots of servers, such that it has an up-time that far exceeds an implementation that uses a file-based persistence strategy?

Third, what if Fedora is built on top of Spark? And the resources exist as "resilient distributed datasets" in which "persistence" is thought of as something really quite different?

Finally, given that Fedora operates with clients entirely through HTTP, it follows that Resources consist of HTTP Bodies or _representations_. What, then, does "persistence fixity" mean? And how does transport-fixity not address the various needs to ensure data integrity for resources?

ajs6f:
I want to speak only to one of @acoburn's questions:

> Second, Fedora is concerned about resource durability, but what does that mean? 

There is a long tradition of defining it very precisely: a resource is durable when its _meaning_ is preserved through time (via semantic transparency and thoughtful, fine-grained content modeling). This is as opposed to having its _form_ preserved through time, which is what is supported by the use of fixity measurements. So the 'D' in Fedora has no relevance here, and whatever color of bucket into which bits are put (RAM, local disk, Spark RDDs elsewhere on the local network, "cloud storage", the akashic records…) doesn't matter for the question of durability.

escowles:
> And what is it about persistence-fixity that cannot be handled by the existing transport-fixity mechanism (e.g. Want-Digest)?

Section 3.5.2: https://fcrepo.github.io/fcrepo-specification/#httpGETLDPNR says:
> GET requests to any LDP-NR must correctly respond to the Want-Digest header defined in [RFC3230] unless the Content-Type of the LDP-NR is a message/external-body extension.

[...]

PR #94, opened 2017-04-19
Requiring HEAD requests to support the same headers as GET requests

escowles:
Requiring HEAD requests to support the same headers as GET requests:
* Want-Digest/Digest headers for LDP-NRs
* Link rel='describes' header for LDP-RS that describes an LDP-NR

Addresses #90

### DIFF
--- a/index.html
+++ b/index.html
@@ -362,6 +362,27 @@
           </p>
         </section>
       </section>
+
+      <section id="httpHEAD">
+        <h2>HTTP HEAD</h2>
+        <p>
+          When the request is to the <a>LDP-RS</a> created to describe a <a>LDP-NR</a>, the response MUST include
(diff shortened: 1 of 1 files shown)


awoods:
@escowles : I completely agree with the sentiment of this PR... but wonder if it would be more comprehensive and concise to simply have a HEAD section that states:
> The HEAD method is identical to GET except that the server MUST NOT
   return a message-body in the response, per https://tools.ietf.org/html/rfc2616#section-9.4

escowles:
@awoods That would be fine with me, and more concise.  Unless anyone objects, I'll update this PR to have that instead of repeating the requirements from the GET section.

zimeon (approved):
I support additional clarification of expectations for HEAD, and I like this very simple form

Issue #102, opened 2017-04-21
Messaging section


acoburn:

ajs6f:
Which specific statements would you like to remove?

On a semi-related note, I think "Messages MUST NOT be emitted until after any changes have been persisted to durable storage." might better read "Messages MUST NOT be emitted until after any changes have been made durable." That's a bit more agnostic to implementation details.

acoburn:
I was thinking that the entire section 7.2 could be reduced to:

> Any operation that causes the state of a resource to durably change must make a notification available with a corresponding resource identifier.

ajs6f:
I wouldn't agree with that. I think that the non-normative note (" Non-normative: consumers of these messages should not expect a strict ordering of the events... ") is very important clarification.

More importantly, I don't think that the remark "Any operation that causes contained resources to change MUST trigger corresponding events for those resources." follows from your changed text, nor does the remark "Any operation on an LDPR that results in a change to the containment or membership triples of a distinct other LDPR MUST also trigger an event for that other LDPR.", nor does "A failed operation (HTTP 4xx or 5xx) MUST NOT emit any messages.", so you are changing the meaning of the spec a good deal.

I would actually be in favor of getting rid of "A failed operation (HTTP 4xx or 5xx) MUST NOT emit any messages." but that's a separate issue.

Changing the other remarks... wouldn't that open the door to a messaging impl that would be not be very usable for indexing? I'm not suggesting that indexing is somehow a benchmark for the spec, just that how much indexing tech would have to change is a good measure in perspective for the size of the change you are suggesting.

acoburn:
A more correct replacement would be:

> Any operation that causes the state of one or more resources to durably change must make available one or more notifications corresponding to the changed resources.

Even the current language falls short on the case of non-contained membership triples changing.

ajs6f:
I'm getting a bit mixed up (which is normal for me). Can you put your more correct replacement replacement in the context of the larger change? How would 7.2 read in entirety?

acoburn:
Honestly, it would be easier to discuss this in the context of a pull request. @awoods gave me the distinct impression that PRs that make substantive changes are not acceptable until after an issue has been discussed to satisfaction. Is that not the case?

ajs6f:
I can't speak for @awoods. I would be very happy to see a PR and I agree that it would make discussion both easier and more fruitful.

acoburn:
Resolved via #106

PR #106, opened 2017-04-21
Clarify text in the Messaging/Notification section

### LINKED ISSUE #105: Should the messaging section be renamed?
If the messaging section is to support RSS feeds, it seems that "messaging" is a bit of a misnomer and that "notifications" would be a better name.

acoburn:


barmintor:
@dannylamb this seems to be mostly nomenclature, and not a decline in specification. Thoughts from the Islandora perspective? @escowles ditto Hydra?

ajs6f (approved):
I'm not sure how strongly asynchrony is implied by "messaging", but I agree that "notification" implies it even less.

escowles (approved):
Most Hydra repositories do not use messaging now, so from the Hydra perspective, this change is unlikely to impact us one way or the other.

I agree with @barmintor that this looks like a nomenclature change, and the requirements for when notifications are generated and what they contain look equivalent, and the text is considerably streamlined too.

awoods:
I will give this a look later today / this evening.

zimeon:
General sense seems fine although I think it would be nice to keep the following two clarifications: 

> A failed operation (HTTP 4xx or 5xx) MUST NOT emit any messages.

> Messages MUST NOT be emitted until after any changes have been persisted to durable storage.

I assume these are taken as implied still, but I prefer making them explicit.

acoburn:
@zimeon:

> Messages MUST NOT be emitted until after any changes have been persisted to durable storage.

implies that an in-memory implementation is not possible. I would vastly prefer to leave that assumption out of the specification.

ajs6f:
> A failed operation (HTTP 4xx or 5xx) MUST NOT emit any messages.

I would prefer to let this go out. An impl might (for audit purposes or similar) prefer to offer the option of (e.g.) notification for 403 or the like.

acoburn:
@zimeon, re:

> A failed operation (HTTP 4xx or 5xx) MUST NOT emit any messages.

do we know enough about the notification use cases such that we don't want to allow this behavior? I can imagine several use cases off the top of my head that might want this.

zimeon:
It seems from comments above that the intention here is to change scope quite significantly (somewhat beyond agreements in #105 and #102), which perhaps needs some discussion (and maybe use cases).

I'm fine with dropping _persisted to durable storage_ down to _persisted_ (in whatever notion of persisted a particular implementation has). However, if we want to allow a much broader scope of events that generate notifications we should say something about that. As currently written the text is all about emitting notifications for resource changes. If there is agreement to extend that to other conditions then there should be changes along the lines of 

> For every resource whose state is changed as a result of an HTTP operation, there MUST be a corresponding notification made available describing that change.

to 

> For every resource whose state is changed as a result of an HTTP operation, there MUST be a corresponding notification made available describing that change. Implementations MAY also emit notifications for other operations.

ajs6f:
@zimeon I would be even more generic: "Implementations MAY also emit notifications for other reasons."

index.html
```diff
@@ -712,101 +712,79 @@ <h2 class='informative'>Vary</h2>
...
-        </p>
-        <p>
-          Messages MUST NOT be emitted until after any changes have been persisted to durable storage.
+          An implementation is free to choose from any transport technology so long as
+          the notifications conform to what is described in the following sections.
         </p>
         <p>
-          Repository start, stop and configuration change operations MAY cause a message to be emitted.
+          Implementers should be aware that some operations cause multiple resources to change.
+          In these cases, there will be a corresponding notification describing each of the changes.
```
awoods, line 731: We should not lose the `MUST` language from:
https://github.com/fcrepo/fcrepo-specification/pull/106/files#diff-eacf331f0ffc35d4b482f1d15a887d3bL753

index.html
```diff
@@ -712,101 +712,79 @@ <h2 class='informative'>Vary</h2>
...
+        <h3>Notification Events</h3>
         <p>
-          A message MUST contain at least one value for the event type. Event types SHOULD be drawn
-          from the [[!activitystreams-vocabulary]].
-        </p>
-        <p>
-          A message SHOULD contain at least one value for the agent operating on the resource.
-        </p>
-        <p>
-          A message SHOULD contain a collection of RDF types corresponding to the resource that
```
awoods, line 781: Since we are deferring to the activitystreams spec, I am not seeing `MUST` language for `event types` (or activity types). Are you intentionally dropping this? If not, can you point to the reference in the activitystreams spec that specifies it msut be included?

awoods (changes requested):

index.html
acoburn: How are we losing the `MUST` language? See this statement:

> For every resource whose state is changed as a result of an HTTP operation, there MUST be a corresponding notification made available describing that change.

ajs6f: I understood @awoods 's comment to be referring to the question of ordering-- that the spec should prevent inferring the ordering of operations from the ordering of notifications. Now I think I must have misunderstood the comment.

index.html
acoburn: The current language is:

> A message must contain at least one value for the event type. Event types should be drawn from the [activitystreams-vocabulary].

The new language is:

>  Wherever possible, data SHOULD be expressed using the [[!activitystreams-vocabulary]].

and

> Each event described by a notification MUST contain ... the event type(s), corresponding to the HTTP operation

How is that dropping the `MUST` language?

awoods: Thanks for pointing the relevant diff out.

index.html
barmintor: I'm sympathetic to @awoods, it's difficult to track the requirement deltas in a text diff (but @acoburn's PR is more legible IMO, and that's the cost) . Here's the removals that I see going over it trying to pull all the norm indicators out:
* "Repository start, stop and configuration change operations MAY cause a noun to be emitted."
* "nouns MAY contain a timestamp marking the time the resource was added/modified/deleted."
* "A failed operation (HTTP 4xx or 5xx) MUST NOT emit any nouns."
* "nouns MUST NOT be emitted until after any changes have been persisted to durable storage."
* Emitted events MUST contain **exactly one value** (cardinality removed) for resource identifier/IRI
* Emitted events MUST contain exactly one value for event identifier

I don't think there are any new requirements in this PR, tho some requirements individually explicated have been collapsed. @acoburn does that line up with your intentions? I think the removal of the event identifier is the only one that surprises me, but because it's in a consolidated set of requirements it may have been an oversight? I can't remember if that was discussed.

acoburn: @barmintor that is a good summary. Here is an explanation for each removal:

> "Repository start, stop and configuration change operations MAY cause a noun to be emitted."

The new (and old) language _requires_ notifications for resource modifications. As long as we don't explicitly forbid other events (which we really shouldn't) then allowing events for start/stop operations is implied and doesn't need to be called out explicitly.

> "nouns MAY contain a timestamp marking the time the resource was added/modified/deleted."

Activity Streams is flexible about the content. It is also true that messages MAY contain any other data, so it is superfluous to state that they MAY contain timestamps.

> "A failed operation (HTTP 4xx or 5xx) MUST NOT emit any nouns."

As per the discussion elsewhere, I don't think we should explicitly _restrict_ certain potential categories of notifications, especially since there are (arguably) use cases for them (e.g. audit)

> "nouns MUST NOT be emitted until after any changes have been persisted to durable storage."

This is removed because it implies a certain type of persistence layer for an implementation, which is inappropriate.

> Emitted events MUST contain exactly one value (cardinality removed) for resource identifier/IRI

Activity Streams allows for a Collection structure (multiple events in a single notification). Any AS/2.0 client can transparently handle these. If we explicitly require one-resource-identifier-per-event, we are implicitly forbidding the use of AS Collections, which would be a terrible thing.

> Emitted events MUST contain exactly one value for event identifier

This is an overreach of the Fedora spec. Transport mechanisms provide their own identifier and in any case a unique identifier may be added by any implementation. The only reason this was ever part of the spec is because the current Fedora4 implementation includes a unique identifier for messages (some sort of UUID). Note -- this is different than the resource identifier, which _is_ required.

acoburn: One additional item here related to:

> Emitted events MUST contain exactly one value (cardinality removed) for resource identifier/IRI

If we are make it possible for other categories of events (e.g. repository start/stop), there may be cases where there will be no resource IRI at all. In effect, I want the specification to allow for events that have 0, 1 or multiple resource IRIs, as that will be most useful for downstream messaging applications.

ajs6f: I have no intention of impling the "atomic ops" spec, but I can imagine impling WebDAV's `MOVE` HTTP message to meet the same use cases. For that message, it would be natural to emit a 2-resource-IRI notification.

acoburn: And using AS/2.0's [Collection](https://www.w3.org/TR/activitystreams-core/#collections) model is a very natural way of doing exactly that. That is precisely the sort of thing that should be allowable.

barmintor: So the language here is around messages changing to notifications, but there is another noun here- event. If a notification can be a collection, is it advisable to still recommend single resource event descriptions within the notification? That seems to be the missing link.

acoburn: @barmintor The suggestion here is to _remove_ language specifying a single-resource-per-notification. The text in question is _removed_ with this PR.

barmintor: @acoburn I understand that with regard to the entireity of the notification, I only want to make sure that there's not a reconciliation with the motivation behind the deleted text at the level of event description, which seems like a second entity.

acoburn: @barmintor at the level of event-description (e.g. within an AS Collection or as a singleton notification) the same rules of description apply, namely that (1) it `MUST` conform to the AS 2.0 spec, (2) if a resource has been created/modified/deleted that IRI `MUST` be included and (3) the event type(s) `MUST` be included. w/r/t a separate "event identifier", the suggestion here is to remove that requirement. In so doing, we do not preclude implementations from minting an identifier for such notifications, but I have yet to see a single use case for event identifiers that would require them to be server-generated. 

Requiring behaviors in the absence of use cases is exactly the sort of thing I'd like to avoid with this specification. Likewise, prohibiting behaviors without a _really_ solid justification is what I also want to avoid. Those two principles are what underlie these changes.

awoods (approved):

zimeon (approved):

PR #109, opened 2017-04-28
Strengthening requirements for respecting PreferInboundReferences Prefer header

### LINKED ISSUE #81: Hydra Observation: Make PreferInboundReferences more strict?
I've been looking at the spec and thinking about Hydra's use cases, mostly about what would happen if we were to use Fedora directly with no Solr synchronization layer. We have members linking to parents, and when we delete members we need to clean out their references from their children. If we don't have PreferInboundReferences, we'd have to do a full Fedora scan to find the "parents."

At the very least, it seems like we should have some way to tell when we ask for PreferInboundReferences that it didn't comply, and we can either scan all of Fedora, ask a secondary index, or just stop using that implementation.

escowles:
Fixes #81

### DIFF
--- a/index.html
+++ b/index.html
@@ -336,9 +336,10 @@
         <section id="additionalPreferValues">
           <h2>Additional values for the <code>Prefer</code> header</h2>
           <p>
-              In addition to the requirements of [[!LDP]], an implementation MAY support either or
-              both of the following additional values for the <code>Prefer</code> header when making
-              <code>GET</code> requests on <a>LDPC</a> resources:
+              In addition to the requirements of [[!LDP]], an implementation MAY support the value
+              <code>http://www.w3.org/ns/oa#PreferContainedDescriptions</code> and SHOULD support the value
+              <code>http://fedora.info/definitions/repository#PreferInboundReferences</code> for the
+              <code>Prefer</code> header when making <code>GET</code> requests on <a>LDPC</a> resources:
           </p>
           <ul>
             <li>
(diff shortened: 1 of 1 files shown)


awoods (approved):

index.html
```diff
@@ -353,6 +354,14 @@
             </li>
           </ul>
         </section>
+        <section id="httpGETLDPRS">
+          <h2><a>LDP-RS</a>s</h2>
+          <p>
+            <code>GET</code> requests that include a <code>Prefer</code> header to any <a>LDP-RS</a> MUST
```
zimeon: Application of the preference is optional (MAY and SHOULD above), so shouldn't this be:

> Responses to <code>GET</code> requests that apply a <code>Prefer</code> request header to any <a>LDP-RS</a> MUST include the <code>Preference-Applied</code> response header as defined in RFC7240 section 3.

zimeon (changes requested):

index.html
awoods: That is a helpful refinement, @zimeon. Changing:
> `GET` requests that include a `Prefer` header...

to:
> Responses to `GET` requests that apply a `Prefer` request header...

..since, regardless of whether the client's `GET` includes the `Prefer` header, this section of the specification only applies if a server applies the `Prefer` header.

escowles: 👍  that is much clearer — I've just pushed a new commit with this updated language.

zimeon (approved):

awoods:
Looks like this is ready to go. Comments @dannylamb ? @barmintor ?

dannylamb:
:+1:

PR #120, opened 2017-05-26
PUT to create becomes MAY, containment undefined

### LINKED ISSUE #43: Revisit PUT for creating new resources

zimeon:
Resolves #43

There seems to be agreement that PUT to create should become MAY which makes this whole section a clarification and thus marked non-normative. Containment of the resulting resource is undefined (following LDP). Expressing constraints via the containing resource ahead of creation (as we mention for POST) doesn't make sense since we don't know whether or what the container might be, so have removed the statement attempting to clarify LDP 4.2.1.6.

### DIFF
--- a/index.html
+++ b/index.html
@@ -320,14 +320,13 @@
           </p>
         </section>
 
-        <section id="httpPUTcreate">
+        <section id="httpPUTcreate" class="informative">
           <h2>Creating resources with HTTP PUT</h2>
           <p>
-            An implementation MUST accept HTTP <code>PUT</code> to create resources.([[!LDP]]
-            <a href="https://www.w3.org/TR/ldp/#ldpr-put-create">4.2.4.6</a> MAY becomes MUST). The default interaction model that will
-            be assigned when there is no explicit <code>Link: rel="type"</code> header in the request MUST be recorded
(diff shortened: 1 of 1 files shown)


acoburn:
@zimeon this looks great, thanks!

escowles:
👍

awoods (approved):

dannylamb (approved):
:shipit:

PR #121, opened 2017-05-28
specify use of message/external-body for referenced content

### LINKED ISSUE #118: Specify a Detectable, Optional Behavior for LDP-NR with message/external-body
A new section of the spec should:

- [ ]  recapitulate the use case for proxied content and justification for message/external-body in an informative preamble
- [ ]  require LDP-C that support the creation of LDP-NR proxying content this way to advertise `message/external-body;type=TYPE` in the `Accept-Post: ` response header for all supported values of TYPE
- [ ]  require LDP-S receiving a PUT or POST request that would create or update a LDP-NR with an unsupported `Content-Type: message/external-body;type=TYPE` value of TYPE to respond with HTTP 415
- [ ]  require LDP-S receiving a PUT or POST request that would create or update a LDP-NR with `Content-Type: message/external-body;type=TYPE` to guarantee that it can provide all GET and HEAD response headers required of LDP-NRs, or reject with a 4xx
- [ ]  LDP-NR GET/HEAD responses SHOULD also include a `Content-Location: ` header with a URI representation of the location of the referenced content.
- [ ] LDP-S SHOULD respect the `expires` parameter on `message/external-body` by either copying the data to other storage or rejecting the request with a 5xx

See also: https://github.com/fcrepo/fcrepo-specification/wiki/External-Content

barmintor:
fixes #118

### DIFF
--- a/index.html
+++ b/index.html
@@ -259,12 +259,12 @@
             A HTTP <code>POST</code> request that includes an unsupported <code>Digest</code> type (as described
             in [[!RFC3230]]), SHOULD be rejected with a 400 Bad Request response.
           </p>
-          <p>
-            Implementations SHOULD support <code>Content-Type: message/external-body</code> extensions for
-            request bodies for HTTP <code>POST</code> that would create <a>LDP-NR</a>s. This content-type requires
-            a complete <code>Content-Type</code> header that includes the location of the external body, e.g
-            <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
-            as defined in [[!RFC2017]].
+          <p class='informative'>
+            Implementations may support <code>Content-Type: message/external-body</code> extensions for
+            request bodies for HTTP <code>POST</code> that would create <a>LDP-NR</a>s.
+            This content-type requires a complete <code>Content-Type</code> header that includes the location
+            of the external body, e.g <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
+            as defined in [[!RFC2017]]. Requirements for this interaction are detailed in <a href="#external-content">External LDP-NR Content</a>.
           </p>
         </section>
       </section>
@@ -294,12 +294,12 @@
             A HTTP <code>PUT</code> request that includes an unsupported <code>Digest</code> type (as described
             in [[!RFC3230]]), SHOULD be rejected with a 400 Bad Request response.
           </p>
-          <p>
-            Implementations MUST support <code>Content-Type: message/external-body</code> extensions
-            for request bodies for HTTP <code>PUT</code> to <a>LDP-NR</a>s. This content-type requires a
-            complete <code>Content-Type</code> header that includes the location of the external body, e.g
-            <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
-            as defined in [[!RFC2017]].
+          <p class='informative'>
+            Implementations may support <code>Content-Type: message/external-body</code> extensions for
+            request bodies for HTTP <code>PUT</code> that would create <a>LDP-NR</a>s.
+            This content-type requires a complete <code>Content-Type</code> header that includes the location
+            of the external body, e.g <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
+            as defined in [[!RFC2017]]. Requirements for this interaction are detailed in <a href="#external-content">External LDP-NR Content</a>.
           </p>
         </section>
 
@@ -372,12 +372,7 @@
           <h2><a>LDP-NR</a>s</h2>
           <p>
             <code>GET</code> requests to any <a>LDP-NR</a> MUST correctly respond to the <code>Want-Digest</code> header
-            defined in [[!RFC3230]] unless the <code>Content-Type</code> of the <a>LDP-NR</a> is a
-            <code>message/external-body</code> extension.
-          </p>
-          <p>
-            <code>GET</code> requests to a <a>LDP-NR</a> with <code>Content-Type: message/external-body</code>, MUST result
-            in an HTTP 3xx redirect message redirecting to the external URL.
+            defined in [[!RFC3230]].
           </p>
         </section>
       </section>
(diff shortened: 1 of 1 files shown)


barmintor:
I'll have limited ability to respond to this thread this week, please feel free to add commits as necessary.

index.html
```diff
@@ -294,12 +294,12 @@
...
             in [[!RFC3230]]), SHOULD be rejected with a 400 Bad Request response.
           </p>
-          <p>
-            Implementations MUST support <code>Content-Type: message/external-body</code> extensions
-            for request bodies for HTTP <code>PUT</code> to <a>LDP-NR</a>s. This content-type requires a
-            complete <code>Content-Type</code> header that includes the location of the external body, e.g
-            <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
-            as defined in [[!RFC2017]].
+          <p class='informative'>
+            Implementations may support <code>Content-Type: message/external-body</code> extensions for
```
ajs6f, line 298: `may` => `MAY`?

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
       </section>
     </section>
 
+    <section id='external-content'>
+      <h3>External LDP-NR Content</h3>
+      <p class='informative'>
+        Variability among client types and locations may mean that LDP-NR
+        content is addressed in ways that are external to LDP-S but not resolvable
```
ajs6f: Where is the def'n for `LDP-S`? What is it?

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+        by all clients. This specification describes the use of <code>message/external-body</code>
+        Content-Type value to signal, on POST or PUT, that the LDP-S should should not consider the
+        request entity to be the LDP-NR's content, but that a Content-Type value will signal a
```
ajs6f: `Content-Type` => `<code>Content-Type</code>`. There are several other missing `<code>` tags.

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+        <p>
+          LDP-S receiving requests that would create or update a LDP-NR
+          with a <code>message/external-body</code> MUST NOT accept the
+          request if it cannot guarantee all of the response headers required
+          by the LDP-NR interaction model in this specification.
+        </p>
+      </section>
+      <section id='external-content-content-location'>
+        <p>
+          LDP-NR GET and HEAD responses SHOULD include a Content-Location: header with a URI
```
ajs6f: Doesn't this destroy the "proxying for authZ" use case?

index.html
acoburn: For non-normative sections (such as this), I think the use of "may" is correct here.

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+          representation of the location of the external content if the LDP-S is proxying
+          the content.
+        </p>
+      </section>
+      <section id='external-content-expires'>
+        <p>
+          Per [[!RFC2017]], all <code>message/external-body</code> Content-Type values
+          may include an <code>expires</code> parameter. LDP-S receiving requests that
```
acoburn: Here, may => MAY

index.html
escowles: Only if the URIs can be resolved by the client, and also doesn't enforce it's own access control.  If either or both of things aren't true, then the server can safely expose the content location URI to the client, even when proxying for authZ.

When not proxying for authZ, exposing the content location URI allows the client to disintermediate the server if it wishes for performance or other reasons.

ajs6f: Exactly-- the server is forced to trust the client (trust that the client won't resolve the URI or that the client will enforce the same authZ), so the use case is unsupported if this `SHOULD` is executed. (It's unusual, to say the least, to make authZ depend on a trustworthy client.) It's not of great interest to me, because my institution does not require this use case, and we will be "implementing" by returning 415s in all cases, so I won't pursue it further.

index.html
awoods: I am guessing `LDP-S` is intended to mean `LDP Server`... which should probably be `the Fedora server`...as `the server` is probably not exactly accurate in this situation.

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+      <section id='external-content-content-location'>
+        <p>
+          LDP-NR GET and HEAD responses SHOULD include a Content-Location: header with a URI
+          representation of the location of the external content if the LDP-S is proxying
+          the content.
+        </p>
+      </section>
+      <section id='external-content-expires'>
+        <p>
+          Per [[!RFC2017]], all <code>message/external-body</code> Content-Type values
```
awoods: I am seeing no mention of `expires` or `expiration` in RFC2017... but do in [RFC1521](https://tools.ietf.org/html/rfc1521)

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+          may include an <code>expires</code> parameter. LDP-S receiving requests that
+          would create or update a LDP-NR with a <code>message/external-body</code> content
+          type SHOULD respect the <code>expires</code> parameter, if present, by copying
+          content. If the expires parameter cannot be accomodated, the request MUST be rejected
```
awoods: typo: `accomodated` should be `accommodated`

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+          with a 4xx or 5xx status code.
```
awoods: A `constrainedby` link header would also be useful here.

index.html
```diff
@@ -889,6 +884,76 @@ <h2 class='informative'>Vary</h2>
...
+        <p>
+          This specification takes no position on the use of <code>message/external-body</code>
+          to create or update LDP-RS with ttl or json-ld.
+        </p>
+        <h4>Proxied Content vs Redirected Content</h4>
+        <p>
+          This specification assumes that clients interested in resolving a <code>type='url'</code>
+          or other value without the LDP-S as an intermediary (effectively, redirection as opposed
+          to proxying) will be able to negotiate the <code>Content-Location</code> response header
+          value from a HEAD request.
```
awoods: Not necessarily a question for the spec, but what would the expected response body be on a GET request of a `redirect`  LDP-NR? ... if we are expecting the client to figure out that an implicit redirect is available via inspection of the `Content-Location` header.

awoods (changes requested):

index.html
zimeon: I agree: s/LDP-S/the Fedora server/

index.html
zimeon: I read this paragraph as specifying a way for a client (at its option) to use `HEAD` to decide whether it can go fetch remote content from the `Content-Location` itself, avoid a default proxy action on `GET`. It seems that from https://tools.ietf.org/html/rfc7231#section-3.1.4.2 the response to `GET` should be either:
  * 200, `Content-Location` and content in body to proxy
  * 3xx, `Content-Location` and `Location` to same URI to redirect (no body)

zimeon:
This PR seems to be heading in the right direction, it would be really helpful to have a pass that fixes the minor formatting issues so that we can focus on the more substantive ones

escowles (approved):

index.html
```diff
@@ -889,6 +884,79 @@ <h2 class='informative'>Vary</h2>
...
+          representation of the location of the external content if the Fedora server is proxying
+          the content.
+        </p>
+      </section>
+      <section id='external-content-expires'>
+        <p>
+          Per [[!RFC1521]], all <code>Content-Type: message/external-body</code> values
+          MAY include an <code>expiration</code> parameter. Fedora servers receiving requests that
+          would create or update a LDP-NR with a <code>message/external-body</code> content
+          type SHOULD respect the <code>expiration</code> parameter, if present, by copying
```
birkland: Is the presence of `expiration` intended to be a mechanism by which a client can instruct the repository to durably persist ephemeral content (see [fedora-tech thread](https://groups.google.com/forum/#!search/fedora$20$20tech/fedora-tech/yzUW8zd2xKc/dNniVeajCQAJ)?

index.html
```diff
@@ -889,6 +884,79 @@ <h2 class='informative'>Vary</h2>
       </section>
     </section>
 
+    <section id='external-content'>
```
awoods: Besides the boilerplate text (Conformance, Terminology, X-Considerations, Acknowledgements, References) the top-level specification elements are the core Fedora capabilities:
1. Resource Management
2. Resource Versioning
3. Binary Resource Fixity
4. Resource Authorization
5. Notifications

With that in mind, I would expect this new section (External LDP-NR Content) to be a sub-element under Resource Management, instead of a new top-level specification element.

index.html
```diff
@@ -889,6 +884,79 @@ <h2 class='informative'>Vary</h2>
...
+      <h3>External LDP-NR Content</h3>
+      <p class='informative'>
```
awoods: This `informative` annotation does not appear to have a visual effect in the rendered spec.

awoods (changes requested):

index.html
birkland: Sorry, I simply find this section hard to understand what this section means from a high level.  It suggests that a server `SHOULD` copy content if an expiration parameter is present, but it could do something else, Including (as the above thread suggests) "expire the resource", which I understand as making the LDP-NR content unavailable once the expiration date is met.  So is it consistent with the spirit of this section that one implementation might interpret `expiration` as "make this ephemeral content durable", and another implementation might interpret it as "expire the content at the appropriate time"?   It may be a stupid question, but I'd really appreciate some help from the editors to help understand how to interpret this section.

awoods: From the MODE implementation, the answer to your [first comment](https://github.com/fcrepo/fcrepo-specification/pull/121#discussion_r120453361) is, yes.
If there is no expectation that this be implemented in any other way, I would suggest we tighten the language to more clearly define what the `expiration` parameter means/requires.

birkland: Great, I believe some tightening and/or non-normative text is necessary if that's the case

awoods (approved):

index.html
```diff
@@ -259,12 +259,12 @@
...
           </p>
-          <p>
-            Implementations SHOULD support <code>Content-Type: message/external-body</code> extensions for
-            request bodies for HTTP <code>POST</code> that would create <a>LDP-NR</a>s. This content-type requires
-            a complete <code>Content-Type</code> header that includes the location of the external body, e.g
-            <code>Content-Type: message/external-body; access-type=URL; URL=\"http://www.example.com/file\"</code>,
-            as defined in [[!RFC2017]].
+          <p class='informative'>
+            Implementations may support <code>Content-Type: message/external-body</code> extensions for
+            request bodies for HTTP <code>POST</code> or <code>PUT</code> that would create <a>LDP-NR</a>s.
```
dannylamb: Do we need to have both verbs here?  Can we just have only POST in the POST section and PUT in the PUT section?

barmintor: If this prevents you from approving, I can get to it sometime this weekend.

dannylamb:
@barmintor I can pull on pull if you're strapped for time.

barmintor:
I wouldn't worry about it, it's the commit-merge-squash-travis dance that
I'll need to wait to do.

dannylamb:
@barmintor ok, nbd

index.html
birkland: Should I create a new issue after this is merged to address this?

barmintor: Yes, please open that ticket.

zimeon (approved):

PR #133, opened 2017-06-14
Restoring persistence fixity language

escowles:


index.html
```diff
@@ -769,6 +769,14 @@ <h2 class='informative'>Vary</h2>
           defined in [[RFC3230]] to <code>POST</code> and <code>PUT</code> requests for <a>LDP-NR</a>s.
         </p>
       </section>
+      <section id="persistence-fixity" class="informative">
+        <h2>Persistence Fixity</h2>
+        <p>
+          Persistence fixity is verified by application of the <code>Want-Digest</code> header defined in
```
acoburn: The language here sounds like this is the _only_ way to implement persistence fixity. And while this is certainly the way I would implement it, perhaps it would be better to use text more like:

> A pattern for verifying persistence fixity of resources may include the application of the <code>Want-Digest</code> ...

ajs6f: I have use cases now for fixity of 3D models for which we could never use this pattern, so there are concrete reasons to be inclusive/expansive.

zimeon: I agree that `Want-Digest` isn't the only way but I'd be a little more positive about it, perhaps:

> Persistence fixity can be verified by application of the <code>Want-Digest</code> header defined in ...

ajs6f: Doesn't that assume that the server impls this semi-essential-optional-core part of the spec?

zimeon:
How about adding to the end of this informative paragraph a back reference to the normative section https://fcrepo.github.io/fcrepo-specification/#httpGETLDPNR ? Maybe an additional sentence:

> See description of [HTTP GET for LDP-NRs](https://fcrepo.github.io/fcrepo-specification/#httpGETLDPNR) and note that [HTTP HEAD](https://fcrepo.github.io/fcrepo-specification/#httpHEAD) is required to return the same headers, thus enabling persistence fixity verification without downloading the content.

birkland:
I think this informative text it a little bit terse; had to re-read that a few times and remember some past conversation to verify that it said what I thought it said.  May I suggest spelling out a little bit more in the text

> `Want-Digest` requires the server to calculate the digest of the response it will return to the client.  In a `GET` request, it is the digest of the response entity body.  In a `HEAD` request, it is the digest of the entity the server would have sent the client if it was a `GET` request.  For LDP-NRs whose content is in persistent storage, `Want-Digest` in a `HEAD` request tells the client the checksum of the persisted content at a given point in time, without having to transfer the entire resource to the user.   A possible pattern for verifying persistence fixity is to store and compare results of `Want-Digest` `HEAD` requests performed at different points in time.

zimeon:
It seems that @birkland and I are thinking along the same lines

acoburn:
> In a GET request, it is the digest of the response entity body

This is _not_ correct. In a `GET` request, the `Digest` header corresponds to the digest of the _resource_ not the response body. Consider the case of a range request where only the first 10 bytes are returned. The `Digest` header, however, will be the same as it is for a non-range request.

acoburn:
I would also like to point out that different server implementations may have very different notions of "persistence". In fact, "persistence" may not even be the right term for an in-memory implementation or an implementation that shards and replicates content across lots of servers.

birkland:
@acoburn   Right.  At a high level, in my mind the explanatory text needs to acknowledge that `Want-Digest` pertains to transmission.  A client can use it to draw additional conclusions (i.e. related to persistence, fixity) only if the client has additional knowledge about the nature of the resource as managed by the server that would support that conclusion. So "For LDP-NRs whose content is in persistent storage.." is an attempt to establish a condition/assumption in a simple way, but maybe it's too subtle.

zimeon (approved):

index.html
```diff
@@ -940,8 +940,18 @@ <h2 class='informative'>Vary</h2>
...
-          Transmission fixity is verified by application of the <code>Digest</code> header
-          defined in [[RFC3230]] to <code>POST</code> and <code>PUT</code> requests for <a>LDP-NR</a>s.
+          Transmission fixity may be verified by including a <code>Digest</code> header (defined in
+          [[RFC3230]]) in <code>POST</code> and <code>PUT</code> requests for <a>LDP-NR</a>s.
+        </p>
+      </section>
+      <section id="persistence-fixity" class="informative">
+        <h2>Persistence Fixity</h2>
+        <p>
+          Persistence fixity may be verified by retrieving the current checksum of an <a>LDP-NR</a> and
```
birkland: @escowles This doesn't address the nuance that `Want-Digest` is inherently about transmission, and drawing the conclusion that it is suitable for persistence fixity requires knowledge of the nature of the LDP-NR.  Can the editors comment about this nuance.  In particular, is the spec presuming that all LDP-NRs are persisted, and therefore by definition getting their checksum is measuring their persistence fixity?  If the spec is not making this presumption, then is the nuance too complicated/subtle to mention?

escowles: @birkland I'm not sure we have to make that inference.  If an in-memory repository can calculate a checksum at multiple points, does it matter if the content is "persisted" or not?

acoburn: If the content isn't necessarily "persisted", then how can the specification make claims about "persistence fixity"?

zimeon: An in-memory implementation still "persists" a resource, just for less time. I think peristence fixity is still a relevant concept

birkland: In any case, I tend to think that an explanation `Want-Digest` that spells out the inference is helpful in understanding what's going on, and how it can be used as a tool to support persistence fixity where applicable.    It also would take a bit of a softer stance on the questions of the nature of persistence.  i.e. if you have a system where you can't infer that `Want-Digest` can be used to support persistence fixity (maybe because the notion of persistence is complex) that's fine.

escowles: @birkland I guess what I'm saying is that the important thing is that the content can be retrieved, and is the same content as some other point in time.  So drawing the connection to the internal state of the repository seems like it's going beyond what the spec should be concerned with, where implementations may reasonably differ.

ajs6f: @escowles I appreciate and share your desire to avoid "going beyond what the spec should be concerned with". I understood the point of @acoburn's https://github.com/fcrepo/fcrepo-specification/pull/133#issuecomment-308475009 to be that it is not possible to do that, by the semantics of `Want-Digest`.

birkland: @escowles Correct, the only thing `Want-Digest` tells you is the checksum of the content of the conveyed resource any given point in time.  Casting it in terms of "can be used to infer persistence fixity" just seems more explicit, clear, and neutral/agnostic about what "persistence fixity" means, for what it's worth.

ajs6f: @zimeon That may be so from a certain POV, but there is a pretty common semantic to the word "persistence" that associates it with some kind of more durable storage (e.g. one that continues across the restart of a node, in "devops" language). Wouldn't it be better to avoid any confusion by avoiding the term?

escowles: @birkland I've pushed another commit that rewords the section with the "can be used to infer persistence fixity", which I agree is clearer and more neutral.  Thanks for bearing with me here!

birkland: @escowles That works for me!  One minor point is that the part about HEAD requests looks like it could stand alone as its own sentence (i.e. not in parens), but that's subjective

birkland:
One thing I just noticed in glancing at RFC7231, section 4.3.2

> The response to a HEAD request is cacheable; a cache MAY use it to
   satisfy subsequent HEAD requests unless otherwise indicated by the
   Cache-Control header field (Section 5.2 of [RFC7234]).  A HEAD
   response might also have an effect on previously cached responses to
   GET;

Obviously if you're using `HEAD` + `Want-Digest` for fixity, you wouldn't want a cached response.  Does the spec need to mention anything about caching as it relates to`Want-Digest` with `HEAD` for performing fixity checks, or is it out of scope?

escowles:
@birkland that's a good point — [RFC 3230](https://tools.ietf.org/html/rfc3230) doesn't say much about that (other than saying a server might be able to reduce checksumming overhead by careful caching).  So I've added a note that the checksums should be calculated instead of cached.

zimeon (approved):
I wonder whether the not-caching note should end up as normative but for now it seems at least good to add a note

escowles:
@zimeon I thought about making it normative, but it seems very hard to test without special knowledge of an implementation that would let you corrupt a binary.

awoods (approved):

barmintor (approved):

dannylamb (approved):

PR #137, opened 2017-06-14
Clarify intent

acoburn:


acoburn:
See post-merge conversation in https://github.com/fcrepo/fcrepo-specification/pull/132

index.html
```diff
@@ -114,11 +114,12 @@
         </li>
       </ul>
       <p>
-        The goal of this specification is to define the behaviors of Fedora server implementations in order to
-        facilitate interoperability with client applications.  Although Fedora servers are not necessarily drop-in replacements
-        for one another, a client developed against this specification should work against any Fedora server with little modification.
-        Some implementations may be unable to provide all features described in this document due to conscious design decisions, but
-        at least the method of rejection for requests against such features is defined.
+        The goal of this specification is to define the fundamental behaviors and interaction
+        patterns of a Fedora server implementation. A consequence of this goal is that client
```
ajs6f: Maybe better just `of a Fedora server.`?

ajs6f (approved):
This is better in that it clearly separates the idea of interoperability and supports it with a promise (`sufficient mechanism to navigate the differences`).

index.html
```diff
@@ -114,11 +114,12 @@
...
       </ul>
       <p>
-        The goal of this specification is to define the behaviors of Fedora server implementations in order to
-        facilitate interoperability with client applications.  Although Fedora servers are not necessarily drop-in replacements
-        for one another, a client developed against this specification should work against any Fedora server with little modification.
-        Some implementations may be unable to provide all features described in this document due to conscious design decisions, but
-        at least the method of rejection for requests against such features is defined.
+        The goal of this specification is to define the fundamental behaviors and interaction
+        patterns of a Fedora server. A consequence of this goal is that client appications will
+        have a well-defined model for interacting with different Fedora implementations.
```
escowles: I think these first two sentences are the core of the disagreement here: I don't agree with them at all, and much prefer the first sentence of the previous wording, which sees specification as a means to the end of interoperability only, not an end in and of itself.

index.html
```diff
@@ -114,11 +114,12 @@
...
+        Client applications should not expect the behaviors of every Fedora server implementation to be
+        identical, but the interaction patterns defined in this specification should provide a
+        sufficient mechanism to navigate the differences between Fedora server implementations.
```
escowles, line 125: I do like this sentence, and think it's a good replacement for the last sentence above.

escowles (changes requested):

index.html
ajs6f: I think you are right, @escowles, that this is the heart of the disagreement. With that in mind, would it not be appropriate to remove discussion of goals from the spec (goals that, it seems clear, are not held in common) and confine the introductory material to discussion of what concrete affordances the spec will provide (i.e. `the interaction patterns defined in this specification should provide...`)?

acoburn: If there is not consensus on this particular language (as @escowles points out), I agree with @ajs6f that it should be removed altogether.

zimeon: See specification charter which defines goals in terms of client-server interop: https://github.com/fcrepo/fcrepo-specification/wiki/Fedora-API-Specification-Charter

ajs6f: None of those goals are incompatible with any of the suggestions that have been made here.

PR #138, opened 2017-06-14
Issue-103 - HEAD must have same headers as GET, except payload header fields

### LINKED ISSUE #103: Is it necessary to clarify HEAD behavior?

zimeon:
Fixes #103. Discussed on API call today: We need HEAD to include `Digest` header so that persistence fixity works and simply strengthening RFC7231's SHOULD to a MUST with the same payload header exemption seems like the best approach.

### DIFF
--- a/index.html
+++ b/index.html
@@ -387,10 +387,14 @@
       <section id="httpHEAD">
         <h2>HTTP HEAD</h2>
         <p>
-         The HEAD method is identical to <code>GET</code> except that the server MUST NOT return a message-body in the response,
-         as specified in [[!RFC7231]] <a href='https://tools.ietf.org/html/rfc7231#section-4.3.2'>section 4.3.2</a>.
-         The server MUST send the same header fields in response to a <code>HEAD</code> request as it would have sent if the
-         request had been a <code>GET</code>, except that the payload header fields (defined in [[!RFC7231]]
+         The <code>HEAD</code> method is identical to <code>GET</code> except that the server
+         MUST NOT return a message-body in the response, as specified in [[!RFC7231]]
+         <a href='https://tools.ietf.org/html/rfc7231#section-4.3.2'>section 4.3.2</a>.
+         The server MUST send the same <code>Digest</code> header in the response as it would
+         have sent if the request had been a <code>GET</code> (or omit it if it would have been
+         omitted for a <code>GET</code>). In other cases, the server SHOULD send the same
+         headers in response to a <code>HEAD</code> request as it would have sent if the
+         request had been a <code>GET</code>, except that the payload headers (defined in [[!RFC7231]]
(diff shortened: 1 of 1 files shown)


ajs6f:
Why was a blanket approach like this considered better than a narrowly-targeted approach (i.e. requiring the operation of `Want-Digest`)?

acoburn:
I will point out that anyone using JAX-RS and Jersey will have a _very_ hard time conforming to this. That includes the current Modeshape-based implementation as well as (probably) any other java-based implementation. Jersey, for instance, will return `Content-Length: 0` on a `HEAD` request, which definitely isn't the same as what is returned with `GET`, and while this language allows that header to be _omitted_ under `HEAD`, it doesn't allow it to be different.

acoburn:
For the reasons described above and in #103 I am 👎  on this change. There are good reasons that RFC7231 uses `SHOULD`. Strengthening it to `MUST` is shortsighted. If the goal is to make sure that `Want-Digest` is supported by `HEAD` requests, I would suggest limiting the text in this specification to that.

zimeon:
@acoburn `Content-Length` is a payload header and thus exempted per https://tools.ietf.org/html/rfc7231#section-3.3 . IMO returning `Content-Lenght: 0` for HEAD when the real length is non-zero is just broken.

ajs6f:
@zimeon https://tools.ietf.org/html/rfc7231#section-4.3.2 makes clear that payload headers may be _omitted_. It does not permit them to be present, but different from what they would be for a `GET`. @acoburn is raising a practical objection.

awoods (approved):

barmintor (approved):

awoods:
I have been doing some experimenting and have determined that in the absence of an application-provided Content-Length header, the servlet container (Jetty, Tomcat, etc) chooses to set the Content-Length or not on HEAD requests.
For example, 
- I use Tomcat8 locally, and it omits the Content-Length header on HEAD
- The UT-Austin demo server (http://demo.fcrepo.org:8080/fcrepo/) uses Tomcat7 and it returns Content-Length=0 on HEAD
- The Jetty server used with Fedora's maven-jetty plugin also returns Content-Length=0 on HEAD

The point here is that there does not appear to be anything we can do at the Fedora level to dictate the servlet container's behavior.
As such, calling out `Digest` is probably the right path forward.

Related Tomcat bug:
https://bz.apache.org/bugzilla/show_bug.cgi?id=59310#c8

zimeon:
Thanks @acoburn & @awoods -- looks we should just follow RFC7231 except for a MUST for `Digest`. Will update the PR tomorrow.

escowles (approved):

zimeon:
I have updated to say MUST only for `Digest`. Although it differs from the writing style of RFC7231 I also changed "header field(s)" to "header(s)" in the three places it occurs in this section, in order to be consistent with the rest of the spec

awoods (approved):

dannylamb (approved):

PR #146, opened 2017-06-25
distinguish use of ldp:contains as non-containment data

### LINKED ISSUE #129: Distinguish ?s ldp:contains ?o triples and containment triples
The LDP spec defines containment triples as:

> A set of triples, maintained by the LDPC, that lists documents **created** by the LDPC but not yet deleted. These triples always have the form: ( LDPC URI, ldp:contains , document-URI ).

(emphasis mine)

However, like squares and rectangles, it's possible to read the definition backwards and observe that while all containment triples have the form `( LDPC URI, ldp:contains , document-URI )`, not all triples of that form are necessarily containment triples. In particular, implementations that separate user and server managed data may be able to support, eg, a DirectContainer that uses `ldp:contains` as its memberRelation.

As such, when the specification refers to `ldp:contains` as the axis along which inheritance of ACLs or other behavior is reckoned, it should be changed to instead refer to the axis of containment.

barmintor:
fixes #129

### DIFF
--- a/index.html
+++ b/index.html
@@ -211,9 +211,23 @@
           <h2>LDP Containers</h2>
           <p>
             Implementations MUST support the creation and management of [[!LDP]] Containers.
-            LDP Direct Containers MUST NOT permit <code>ldp:contains</code> as their membership-predicate
-            and requests that would do so MUST fail with 409 Conflict.
+          </p>
+          <p>
+            LDP Containers MUST distinguish containment triples from membership and minimal-container triples.
+            If an implementation cannot make this distinction solely from the predicate, it MUST NOT permit
+            <code>ldp:contains</code> as the <code>membership predicate</code> for LDP Direct and Indirect
+            Containers or as a client-managed property, and requests that would do so MUST fail with 409 Conflict
+            and accompanying constraints document. Implementations that can distinguish containment triples MAY
+            permit  <code>ldp:contains</code> as a <code>membership predicate</code> or client-managed property;
+            such implementations SHOULD support <code>Prefer:</code> header values allowing clients to distinguish this data in the representation of a LDP-R as defined in [[!LDP]]:
+            <code>http://www.w3.org/ns/ldp#PreferContainment</code>,
+            <code>http://www.w3.org/ns/ldp#PreferMembership</code>, and
+            <code>http://www.w3.org/ns/ldp#PreferMinimalContainer</code>.
             ([[!LDP]] <a href="https://www.w3.org/TR/ldp/#ldpdc-containtriples">5.4.1.4 expansion</a>)
+          </p>
+          <p class='informative'>
+            Non-normative note: [[!LDP]] defines containment relationships as binding an LDPC to LDPRs whose lifecycle it controls and is aware of, and are expressed through containment triples of the form &lt;LDPC-URI&gt; ldp:contains &lt;LDPR-URI&gt;. However, implementations that segregate containment
+            and membership triples may allow the creation of other triples with the ldp:contains predicate, and thus triples with this predicate do not necessarily imply containment.
           </p>
         </section>
       </section>
@@ -423,7 +437,7 @@
           value with a 400 (Bad Request).  Implementations MUST define their own default behavior if a Depth header is not present.
           </p>
           <p>
-          If a server supports recursive <code>DELETE</code>, that recursion MUST be reckoned along the <code>ldp:contains</code> axis linking contained resources.
+          If a server supports recursive <code>DELETE</code>, that recursion MUST be reckoned along the [[!LDP]] <code>containment relationships</code> linking contained resources.
           </p>
         </section>
       </section>
@@ -812,7 +826,7 @@
         <h2>Inheritance</h2>
         <p>
           Inheritance of ACLs as defined in [[!SOLIDWEBAC]] MUST be
-          reckoned along the <code>ldp:contains</code> axis linking
+          reckoned along the {{!LDP]] <code>containment</code> relationships linking
           controlled resources.
         </p>
       </section>
--- a/index.html
+++ b/index.html
@@ -213,12 +213,12 @@
             Implementations MUST support the creation and management of [[!LDP]] Containers.
           </p>
           <p>
-            LDP Containers MUST distinguish containment triples from membership and minimal-container triples.
+            LDP Containers MUST distinguish [<a href="https://www.w3.org/TR/ldp/#dfn-containment-triples">containment triples</a>] from [<a href="https://www.w3.org/TR/ldp/#dfn-membership-triples">membership</a>] and [<a href="https://www.w3.org/TR/ldp/#dfn-minimal-container-triples">minimal-container</a> triples.
             If an implementation cannot make this distinction, it MUST NOT permit <code>ldp:contains</code> as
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -812,7 +826,7 @@ <h2 class='informative'>Vary</h2>
         <h2>Inheritance</h2>
         <p>
           Inheritance of ACLs as defined in [[!SOLIDWEBAC]] MUST be
-          reckoned along the <code>ldp:contains</code> axis linking
+          reckoned along the {{!LDP]] <code>containment</code> relationships linking
```
awoods: Looks like a typo: `{{!LDP]]`

index.html
```diff
@@ -211,10 +211,24 @@
           <h2>LDP Containers</h2>
           <p>
             Implementations MUST support the creation and management of [[!LDP]] Containers.
-            LDP Direct Containers MUST NOT permit <code>ldp:contains</code> as their membership-predicate
-            and requests that would do so MUST fail with 409 Conflict.
+          </p>
+          <p>
+            LDP Containers MUST distinguish containment triples from membership and minimal-container triples.
+            If an implementation cannot make this distinction solely from the predicate, it MUST NOT permit
```
awoods: > If an implementation cannot make this distinction solely from the predicate

For example, are you saying here that if `ldp:contains` is used as a membership predicate in some cases (maybe because it is being used as an `ldp:hasMemberRelation`), then the distinction of whether `ldp:contains` is a containment triple or a membership triple based on the predicate alone is not possible... and therefore, `ldp:contains` can not actually be used as a membership predicate??

index.html
```diff
@@ -211,10 +211,24 @@
...
+            <code>ldp:contains</code> as the <code>membership predicate</code> for LDP Direct and Indirect
+            Containers or as a client-managed property, and requests that would do so MUST fail with 409 Conflict
+            and accompanying constraints document. Implementations that can distinguish containment triples MAY
```
awoods: Containment triples are always of the form: `LDPC URI, ldp:contains , document-URI`
* https://www.w3.org/TR/ldp/#dfn-containment-triples

An example scenario of how containment triples could be distinguished from other `ldp:contains` triples would be helpful.

index.html
ajs6f: Why `solely from the predicate`? Any means of making the distinction works for your "testability" condition.

index.html
```diff
@@ -211,10 +211,24 @@
...
+            permit  <code>ldp:contains</code> as a <code>membership predicate</code> or client-managed property;
+            such implementations SHOULD support <code>Prefer:</code> header values allowing clients to distinguish this data in the representation of a LDP-R as defined in [[!LDP]]:
```
awoods: `such implementations SHOULD support` should probably become `MUST support`, no?

index.html
```diff
@@ -211,10 +211,24 @@
...
+            <code>http://www.w3.org/ns/ldp#PreferContainment</code>,
+            <code>http://www.w3.org/ns/ldp#PreferMembership</code>, and
+            <code>http://www.w3.org/ns/ldp#PreferMinimalContainer</code>.
             ([[!LDP]] <a href="https://www.w3.org/TR/ldp/#ldpdc-containtriples">5.4.1.4 expansion</a>)
           </p>
+          <p class='informative'>
+            Non-normative note: [[!LDP]] defines containment relationships as binding an LDPC to LDPRs whose lifecycle it controls and is aware of, and are expressed through containment triples of the form &lt;LDPC-URI&gt; ldp:contains &lt;LDPR-URI&gt;. However, implementations that segregate containment
+            and membership triples may allow the creation of other triples with the ldp:contains predicate, and thus triples with this predicate do not necessarily imply containment.
```
awoods: Minor: Let's wrap lines at a consistent width.

awoods (changes requested):

index.html
barmintor: That's a fair point, the phrase isn't doing anything and the mechanism for such distinctions is not relevant to the client of the testability of the requirement.

index.html
barmintor: I'm open to that, but I'd like to think about it more. Do you feel like that decision is necessary to make on this PR?

index.html
barmintor: Maybe they're stored in a different graph? It would depend on the implementation. The critical point is that they're either distinguished or disallowed, but I'm happy to add some more text to the informative note.

ajs6f: Easy to imagine an impl that stores LDP-containment info centrally but distributes other data. Two different stores, easy to distinguish. Immediately consistent answers for many LDP requests, but eventual consistency for others.

index.html
```diff
@@ -214,15 +214,14 @@
...
-            permit  <code>ldp:contains</code> as a <code>membership predicate</code> or client-managed property;
-            such implementations SHOULD support <code>Prefer:</code> header values allowing clients to distinguish this data in the representation of a LDP-R as defined in [[!LDP]]:
-            <code>http://www.w3.org/ns/ldp#PreferContainment</code>,
-            <code>http://www.w3.org/ns/ldp#PreferMembership</code>, and
-            <code>http://www.w3.org/ns/ldp#PreferMinimalContainer</code>.
+            If an implementation cannot make this distinction, it MUST NOT permit <code>ldp:contains</code> as
+            the <code>membership predicate</code> for LDP Direct and Indirect Containers or as a client-managed
+            property, and requests that would do so MUST fail with 409 Conflict and accompanying constraints
+            document. Implementations that can distinguish containment triples MAY permit <code>ldp:contains</code>
+            as a <code>membership predicate</code> or client-managed property; such implementations SHOULD support
```
awoods: This all looks good to me, with the exception of `such implementations SHOULD support...`.
It seems like these `Prefer` headers provide the mechanism for the user to distinguish containment from membership triples. 
Since much of the content of this PR is focused on being able to tease apart containment and membership triples, making that functionality available to users seems like a `MUST`.

escowles: Could an implementation use named graphs or some other mechanism to separate the user-provided triples from the containment triples?  If we think there is some other mechanism to do this, then I'm fine with leaving this as SHOULD.

ajs6f: Suppose an implementation _can_ distinguish such classes of triples (by some such means as has been discussed at several points on this PR) but chooses not to provide that ability to clients. Why is that dangerous? Why should it be forbidden?

escowles: @ajs6f Because the client would be able to modify some triples but not others, with no way to tell the difference but trial and error.

ajs6f: @escowles By hypothesis the server does not offer any `ldp:contains` triples that are other than containment triples (to any client). A client decides to test that. The server turns out to be quite certain about its intentions and limitations.

This seems to me to be a question of providing good responses to `Prefer` headers that aren't supported-- not requiring more support for more `Prefer` headers...

awoods (approved):

escowles (approved):

zimeon (approved):

acoburn (commented):
Could you please add line breaks so that the lines wrap at a sensible length?

dannylamb (approved):

escowles:
All the editors have signed off, so I'm going to merge this.  @acoburn there are other line-length issues in this document, so I'll send a follow-on PR to consistently indent the whole doc.

PR #147, opened 2017-06-29
Reflowing line lengths to be 120 characters or less.

escowles:
Non-functional update: reflowing text to be 120 characters or less per line.  In my browser (with generous font sizes), none of the new text is cut off in the diff display.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,6 @@ jdk:
 install:
  - pip install --user html5validator
 
-script: html5validator
-
+script:
+ - html5validator
+ - test `cut -c121- index.html | grep -vx "^$" | wc -l` -lt 2
--- a/index.html
+++ b/index.html
@@ -97,14 +97,14 @@
     <section id='abstract' class='informative'>
       <h2>Introduction</h2>
       <p>
-        This specification refines the semantics and interaction patterns of [[LDP]] in order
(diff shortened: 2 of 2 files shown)


acoburn:
Thanks @escowles! 👍

awoods:
It could make sense to have this ensured by Travis:
https://github.com/w3c/webrtc-respec-ci

escowles:
@awoods That looks like a good thing to add to our CI — though I don't see any MacOSX support, so it's a little hard for me to test.

awoods:
@escowles : I suspect there are other options... maybe:
https://perlmaven.com/run-perl-tidy-to-beautify-the-code

escowles:
Though if we just want to check for 120+ character lines (other than the overrideCopyright one), we could just do:

```
script:
- test `cut -c121- index.html | grep -vx "^$" | wc -l` -lt 2
```

zimeon:
:+1: on these changes, :+1: on adding the 120-char travis check. I think it would also be good to have a scripted auto-format provided we can find something that isn't too ugly.

escowles:
I've added the simple test for whether there are lines longer than 120-chars.  I agree it would be nice to have an autoformatter, though i'm not sure I like the output of tidy.  Does anybody have any other suggestions?

barmintor:
👍

PR #153, opened 2017-07-10
Add missing <section> tag for 3.8.2

zimeon:
Add missing \<section> tag so that 'Proxied Content vs. Redirected Content' gets a section number 3.8.2. Currently the \<h4> after [3.8.1](https://fcrepo.github.io/fcrepo-specification/#external-content-caveats) has no section number.

### DIFF
--- a/index.html
+++ b/index.html
@@ -507,6 +507,8 @@
             This specification takes no position on the use of <code>message/external-body</code> to create or update
           LDP-RS with ttl or json-ld.
           </p>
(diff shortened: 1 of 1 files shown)


awoods (approved):

awoods:
I would suggest if one more @fcrepo/editors is ok with this PR, that they just merge it.
This is really a formatting issue.

escowles (approved):

Issue #148, opened 2017-06-30
Checksum calculation on GET


awead:
In section 3.5.3:

> Given the use of checksums to validate content integrity, the digest values should be calculated instead of cached.

This seems to imply that every GET request to a non-RDF resource should include a checksum calculation. Are we saying it's better to do this all the time, but you don't have to?  I'm thinking of large binary resources and the resulting response time.

escowles:
I was thinking that the ETag could use a cached value, but when the client sent a `Want-Digest` header, that would trigger calculating the checksum instead of using the cached value.

But this note definitely doesn't say that, so making that explicit would be good.

awead:
@escowles thanks, yeah, I just read the section about `Want-Digest`. You could re-state something to that effect here if that helps to clarify it.

awoods:
@awead : If you have time for creating a PR adding such text, that would be very appreciated.

zimeon:
@awead : If you are in a PR mood you might also do one to add yourself as a contributor

awead:
@zimeon thanks, I'm honored!

awoods:
Resolved with: https://github.com/fcrepo/fcrepo-specification/pull/155

PR #155, opened 2017-07-10
Clarify checksum calculation in LDP-NR GET requests, ref #148

awead:
Previous language was unclear if checksums were to be calculated at every GET request, as opposed to only when the Want-Digest was provided.

### DIFF
--- a/index.html
+++ b/index.html
@@ -413,8 +413,9 @@
             header defined in [[!RFC3230]].
           </p>
           <p class='informative'>
-            Non-normative note: Given the use of checksums to validate content integrity, the digest values should be
-            calculated instead of cached.
+            Non-normative note: In the presence of a <code>Want-Digest</code> header, the current <code>Digest</code>
+            value should be calculated to enable fixity verification. Otherwise, it is expected that ETag values may be
+            cached. See <a href="#persistence-fixity"></a>.
           </p>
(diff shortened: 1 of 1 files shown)


zimeon:
The new text seems to imply that the digest header will be returned even if not requested (which is allowed but surely not expected). I don't know whether we want to mention ETag specifically but, as discussed in #148, it seems that is the thing we expect to be cached (and must be present per LDP).  Perhaps:

> Non-normative note: In the presence of a `Want-Digest` header, the current `Digest` value should be calculated to enable <a href="#persistence-fixity">fixity verification</a>. Otherwise it is expected that `ETag` values may be cached.

awead:
@zimeon works for me. Thought, @escowles ? As to ETags or checksums git GET requests to LDP-NRs *without* any headers, that could be a separate ticket.

escowles:
Yes, I agree: I would only expect a `Digest` header in response to a `Want-Digest` header, and using the cached value for the `ETag` is reasonable.

awead:
@zimeon It's been amended.

zimeon (approved):

awoods (approved):

dannylamb (approved):

escowles (approved):

Issue #152, opened 2017-07-10
Specification should reference local respec javascript


awoods:
Due to the unpredictable changes to:
https://www.w3.org/Tools/respec/respec-w3c-common

..the specification should manage the rate of change by referencing a local copy of the respec javascript.

zimeon:
:+1: -- makes sense not to have it break on us unexpectedly (again)

PR #158, opened 2017-07-10
Add local version of respec

awoods:
- respec v14.0.13

Resolves: https://github.com/fcrepo/fcrepo-specification/issues/152


index.html
```diff
@@ -4,7 +4,7 @@
     <title>Fedora API Specification</title>
     <meta charset='utf-8'>
     <meta name="description" content="Fedora API Specification">
-    <script src='https://www.w3.org/Tools/respec/respec-w3c-common'
+    <script src='http://fedora.info/spec/respec/v14.0.13/respec-w3c-common'
```
acoburn: Please use `//fedora.info/spec...` (omitting http:), as this change may cause many browsers to _not_ load the javascript if the page is served via HTTPS. Using `src="//fedora.info...` will tell browsers to use the same protocol (HTTP or HTTPS) as is used by the main page.

acoburn: Obviously, this depends first making sure that the server at `fedora.info` supports HTTPS connections; otherwise, you would be better off just using a relative link (omitting the hostname) to the javascript file being added to this repository.

zimeon: I'd go with a relative link. That way it will work with a locally checked out copy of the spec repo which I know I find convenient.

acoburn: A relative link would be very easy given that the JS file is added to the repo with this PR.

awoods: I initially had a relative link in the same directory as the index.html file. The reason a backed away from that was because browsing to the local index.html in the browser with `file://` did not load the javascript.
But if others are ok with that limitation, I am happy to update.

acoburn: I think the _best_ solution is to add HTTPS support to `fedora.info`, put the javascript file there and link to it using HTTPS. The downside of this is that it requires more server-side setup; the upside is that `fedora.info` _should_ use HTTPS anyway.

The _easiest_ solution is to use a relative link. The downside is, as @awoods described, there may be issues loading the javascript (I have not ever encountered this issue with Chrome or Firefox and cannot reproduce such an error using the `file://` protocol); the upside is that it requires no server-side changes.

Either way is fine with me -- it's the current text in this PR that isn't OK (it would cause the editor's draft to be considerably less usable).

awoods: The local `file://` relative issue I described goes away if adding the `.js` suffix to `respec-w3c-common`.

acoburn:
Thanks, @awoods. That's much better.

zimeon (approved):
Looks good (and I got a little education because I had to look up the [`class=remove`](https://github.com/w3c/respec/wiki/remove))

escowles (approved):

dannylamb (approved):

PR #160, opened 2017-07-12
Switch sections 3.4.1 and 3.4.2 to match order for GET

zimeon:
No change to content. Just switch LDP-NR and LDP-RS sections 3.4.1 / 3.4.2 in PUT to match order for HTTP GET method and LDP spec.


awoods:
As this is a simple format issue, I would suggest if we have one more editor who approves this change... we merge it.

awoods (approved):

escowles (approved):

PR #162, opened 2017-07-12
Add section for current 3.1 (intro) and clarify LDP-NR creation rule

zimeon:
This PR does *not* intend to change meaning. It:

  1. Makes a sub-section for the text that is currently the intro for [3.1 General](https://fcrepo.github.io/fcrepo-specification/#general)
  2. Gives context for the statement about LDR-NR creation (while avoiding copying any of the RFC terms from LDP 5.2.3.4) -- reading it had tripped me up a couple of times


awoods (approved):

escowles (approved):

barmintor:
How would you feel about adding "regardless of Content-Type: value"?

zimeon:
@barmintor - :+1: , I think that is a helpful clarification. It is a little awkward to add in the text because I think putting it at the end of sentence badly separates the clause from the rest of the description of resource creation ... I think in the parens is the best option without a more major rewrite. Will update PR

awoods (approved):

barmintor (approved):

PR #173, opened 2017-07-19
Add notion of default ACL, every controlled resource must have ACL

### LINKED ISSUE #163: Reconcile WAC and the possibility of uncontained resources in Fedora
The Web Access Control spec relies upon [all documents being contained](https://github.com/solid/web-access-control-spec#containers-and-inherited-acls) and [every root container must have an ACL resource](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm). The Fedora API spec allows for uncontained resources which creates situations where the WAC rules don't work. Options I can see include:

  * Say that every uncontained resource (that isn't itself an ACL) MUST have an ACL resource (implication for DELETE debate: if a DELETE zapped a parent the server would need to propagate ACL down to orphans)
  * Say that at the end of the WAC inheritance process some default should be used if no ACL was found (question about how that default is configured or set/adjusted)

zimeon:
Fixes #163

### DIFF
--- a/index.html
+++ b/index.html
@@ -838,7 +838,9 @@
         <p>
           A conforming server MUST advertise the appropriate Access Control List Resource (ACL) for every controlled
           resource in HTTP responses with a <code>Link: rel="acl"</code> header. This interaction pattern is described
-          in [[!SOLIDWEBAC]].
+          in [[!SOLIDWEBAC]]
+          <a href="https://github.com/solid/web-access-control-spec#acl-resource-location-discovery">ACL Resource
+          Location Discovery</a>.
         </p>
       </section>
       <section id="inheritance">
--- a/index.html
+++ b/index.html
@@ -829,24 +829,29 @@
       <section id="solid-ldp-acls">
         <h2>ACLs are LDP RDF Sources</h2>
         <p>
-          The linked Access Control List Resource (ACL) for a controlled resource by a conforming server MUST itself be
+          The Access Control List Resource (ACL) for a controlled resource on conforming server MUST itself be
           an LDPRS. This ACL resource SHOULD be located in the same server as the controlled resource.
         </p>
       </section>
       <section id="link-rel-acl">
         <h2>ACLs are discoverable via Link Headers</h2>
         <p>
-          A conforming server MUST advertise the linked Access Control List Resource (ACL) for a controlled resource in
-          HTTP responses with a <code>Link: rel="acl"</code> header. This interaction pattern is described in [[!WEBAC]]
-          and [[!SOLIDWEBAC]].
+          A conforming server MUST advertise the appropriate Access Control List Resource (ACL) for every controlled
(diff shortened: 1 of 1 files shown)


awoods:
Minor Travis issue with line length.

zimeon:
Thanks @awoods - was in process of fixing, and now fixed

index.html
```diff
@@ -829,24 +829,29 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
...
       </section>
       <section id="link-rel-acl">
         <h2>ACLs are discoverable via Link Headers</h2>
         <p>
-          A conforming server MUST advertise the linked Access Control List Resource (ACL) for a controlled resource in
-          HTTP responses with a <code>Link: rel="acl"</code> header. This interaction pattern is described in [[!WEBAC]]
-          and [[!SOLIDWEBAC]].
+          A conforming server MUST advertise the appropriate Access Control List Resource (ACL) for every controlled
+          resource in HTTP responses with a <code>Link: rel="acl"</code> header. This interaction pattern is described
+          in [[!SOLIDWEBAC]].
```
awoods: Recommend adding a direct link to: https://github.com/solid/web-access-control-spec#acl-resource-location-discovery

awoods (changes requested):

index.html
zimeon: Good call, will update PR

index.html
```diff
@@ -829,24 +829,31 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
       <section id="solid-ldp-acls">
         <h2>ACLs are LDP RDF Sources</h2>
         <p>
-          The linked Access Control List Resource (ACL) for a controlled resource by a conforming server MUST itself be
+          The Access Control List Resource (ACL) for a controlled resource on conforming server MUST itself be
```
escowles: Minor edit: can you change "on conforming server" to "on a conforming server"?

escowles (approved):
👍  Other than a very minor edit, this looks good to me.

index.html
zimeon: Thanks @escowles, will update

awoods (approved):

acoburn:
One question about the default ACL and the corresponding link header. If a server `MUST` supply a default ACL for controlled resources that don't have an ACL defined in the relevant hierarchy, if follows that the server `MUST` also provide a link to that ACL in the HTTP response headers. Does that imply that the default ACL must be dereferencable?

zimeon:
@acoburn - yes, I think so. I wonder whether we should add that explicitly (it is testable)?

acoburn:
@zimeon thanks! (n.b. this is another difference w/ the current Fedora WebAC module, but that's entirely OK from my perspective).

zimeon:
Have created #175 to separate the decision about saying default MUST be resolvable

dannylamb (approved):

barmintor (approved):

PR #181, opened 2017-07-26
Change external-body 'type' to 'access-type'

awoods:
..per https://tools.ietf.org/html/rfc1521#section-7.3.3

### DIFF
--- a/index.html
+++ b/index.html
@@ -483,19 +483,20 @@
           the use of <code>Content-Type: message/external-body</code> values to signal, on POST or PUT, that the Fedora
           server should should not consider the request entity to be the LDP-NR's content, but that a
           <code>Content-Type</code> value will signal a name or address at which the content might be retrieved. The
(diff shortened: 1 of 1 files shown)


barmintor (approved):
Thanks for catching this mistake, @awoods.

escowles (approved):

zimeon (approved):

zimeon:
This change is good... but, related, is there are reason why we cite https://tools.ietf.org/html/rfc1521#section-7.3.3 instead of https://tools.ietf.org/html/rfc2046#section-5.2.3 that obsoletes it? (might also be nice to add the direct link to the section of the RFC after the normative bibliography link)

awoods:
@zimeon : I will create a follow-on PR for updating the referenced spec and adding a direct link, so as not to hold up this PR.

zimeon:
:+1:

PR #189, opened 2017-08-10
Add fcrepo:VersionedResource to indicate versioning

### LINKED ISSUE #130: Clarification of resource version creation

zimeon:
Fixes #130 

Also changes related instances of "LDP Resource" and "LDP Container" to linked "LDPR" and "LDPC" respectively, to be consistent (1 each).

### DIFF
--- a/fcrepo.ttl
+++ b/fcrepo.ttl
@@ -21,3 +21,8 @@
     rdfs:isDefinedBy : ;
     vs:term_status "working-draft" .
 
+:VersionedResource a rdfs:Resource ;
+    rdfs:label "versioned resource"@en ;
+    rdfs:comment "An IRI to signal that a versioned resource should be created."@en ;
+    rdfs:isDefinedBy : ;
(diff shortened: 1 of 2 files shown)


escowles (dismissed):

fcrepo.ttl
```diff
@@ -21,3 +21,8 @@
     rdfs:isDefinedBy : ;
     vs:term_status "working-draft" .
 
+:VersionedResource a rdfs:Resource ;
```
acoburn: Shouldn't this be `rdfs:Class` instead of `rdfs:Resource`? If this is a "type" that a resource (e.g. the LDPCv) would inherit, then it seems like this is more of a class than an individual. And if this is an "interaction model" similar to the LDP interaction models, that would also argue for making this a class.

zimeon: @acoburn - I agree, thanks, will change

awoods (approved):

escowles (approved):

PR #191, opened 2017-08-10
Minor formatting fix

zimeon:
No (intended) change to content, just missing term links and \<code>

### DIFF
--- a/index.html
+++ b/index.html
@@ -243,12 +243,12 @@
             ([[!LDP]] <a href="https://www.w3.org/TR/ldp/#ldpdc-containtriples">5.4.1.4 expansion</a>)
           </p>
           <blockquote class="informative">
-            Non-normative note: [[!LDP]] defines containment relationships as binding an LDPC to LDPRs whose lifecycle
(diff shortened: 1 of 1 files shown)


awoods (approved):
As this is only a formatting change, I would suggest the next @fcrepo/editors to review this PR should merge it.

acoburn:
While I agree with @awoods' general intent here, his suggestion clearly violates the policy outlined here: https://github.com/fcrepo/fcrepo-specification/wiki/Issue-Resolution-Process

As such, I am 👎  on having "the next @fcrepo/editors to review this PR" to merge this. Instead, I would strongly suggest that the process document be changed to accommodate these sorts of situations. Otherwise, what is the point of the process document if you can just ignore it whenever it is convenient? 

As an alternative, I would encourage you to make a distinction between "substantive" changes and "non-substantive" changes (i.e. in the text of the process document). I would expect that any PR that changes normative language would automatically be "substantive", and these sorts of PRs would be "non-substantive". This would give the editors some level of flexibility for "minor edits".

escowles (approved):

Issue #175, opened 2017-07-20
Should we say default ACL MUST be resolvable?


zimeon:
Raised in discussion of #173 with @acoburn:

> If a server MUST supply a default ACL for controlled resources that don't have an ACL defined in the relevant hierarchy, it follows that the server MUST also provide a link to that ACL in the HTTP response headers. Does that imply that the default ACL must be dereferencable?

I think the answer is yes and it seems that this is testable so would be OK to say _MUST_.

acoburn:
One consideration here is the question: how would this differ from or clarify the existing SOLIDWEBAC statement that the root container must have a configured ACL.

dannylamb:
By virtue of an ACL being an LDP-RS, does that not imply it is resolvable?

zimeon:
Given discussion on #198 I no longer agree with my original https://github.com/fcrepo/fcrepo-specification/issues/175#issue-244418219 that ACL _MUST_ be resolvable. However, given the different interpretation of WAC on this issue I think we need to clarify that `rel="acl"` will always point the individual resource ACL (and not one found by inheritance). I have an idea for how to do this so will make a PR to discuss the text.

PR #198, opened 2017-08-21
Target of ACL Link header must be resolvable

awoods:
Resolves https://github.com/fcrepo/fcrepo-specification/issues/175

### DIFF
--- a/index.html
+++ b/index.html
@@ -861,8 +861,10 @@
         <h2>ACLs are discoverable via Link Headers</h2>
         <p>
           A conforming server MUST advertise the appropriate <a>ACL</a> for every controlled resource in HTTP
-          responses with a <code>Link: rel="acl"</code> header. This interaction pattern is described in
-          [[!SOLIDWEBAC]]
+          responses with a <code>Link: rel="acl"</code> header.
+          Per [[!RFC5988]] <a href="https://tools.ietf.org/html/rfc5988#section-5.1">section 5.1</a>, the target IRI
+          of the <code>Link</code> header must be resolvable.
+          This interaction pattern is described in [[!SOLIDWEBAC]]
           <a href="https://github.com/solid/web-access-control-spec#acl-resource-location-discovery">ACL Resource
           Location Discovery</a>.
         </p>


acoburn:
The Solid WAC document states (in its inheritance algorithm):

> 1. Use the document's own ACL resource if it exists (in which case, stop here).

A resource's own ACL document is described as:

> Given a URL for an individual resource or container, a client can discover the location of its corresponding ACL by performing a HEAD (or a GET) request and parsing the rel="acl" link relation.

This implies that a document's own ACL resource may not exist. And the implication of that is that such a non-existent ACL document may not be resolvable.

This is different from how the current Fedora4 WAC module works (where the rel="acl" link points to the "effective ACL" rather that a resource's own ACL document).

It strikes me that the addition of the language in this PR will present a number of difficulties for implementations. It certainly makes it difficult for my implementation. As such I am 👎  on this PR.

acoburn:
As a note: the reason I initially brought up this issue relates to the fact that the Fedora4 WebAC module has a non-resolvable default ACL resource on the repository root (i.e. `info:fedora/...`). As @dannylamb already [mentioned](https://github.com/fcrepo/fcrepo-specification/issues/175#issuecomment-319761387), the fact that an ACL resource must be an LDP-RS already addresses my concern.

zimeon:
@acoburn - I read WAC differently than you. My sense of the WAC spec statement _"Use the document's own ACL resource if it exists"_ is that either there is an ACL resource (with a URI that we might insist resolves) or there isn't an ACL resource (and therefore no URI, so resolution is a meaningless question). In this second case one uses the inheritance algorithm to find an ACL and this effective ACL is what goes in the link header (as now in F4, except that with this PR we demand that it be resolvable). Can you explain what makes this difficult to implement?

acoburn:
@zimeon what I'm saying is that in the SOLID _reference implementation of WebAC_, you can have a resource, e.g. "/repository/resource" with a link header pointing to it's ACL resource (e.g. "/repository/resource.acl") but that ACL resource returns a 404 (i.e. it's not "dereferenceable"). That behavior makes perfect sense to me as someone implementing WebAC -- but this PR seems to be explicitly departing from that behavior. Again, I am strongly 👎  on this PR.

birkland:
OK, so it seems like the weight of this PR comes from section 5.2, which states that each resource MUST have an acl link. 

As a client, a 404 tells me that there isn't a specific ACL for a resource, and that the defaults apply.  If I want to create an ACL, then I know where to `PUT` it.  If I'm interested in calculating the effective ACL for some reason, then I can apply the algorithm (which is guaranteed to produce a deterministic result that agrees with the repository... right?).  It may be expensive, but I can do it.

If the acl link resolves to a generated "effective ACL", then it's easy for a client to get the default ACL.  However, I might not be able to tell that it's an effective-acl.   It may not matter, though.  Can I `PATCH` or `PUT` it, and result in the expected behavior?  I hope so.  If so, great!

Either one sounds reasonable from my perspective as a client.  As such, I'd be OK with this being an implementation decision.   That means _not_ applying this PR, and leaving the existing language.

acoburn:
@birkland the problem with the semantics of "effective ACL" (e.g. in the current F4 implementation) is that a client really can't easily tell which ACL resource to modify. For instance, given this setup:

    /root
    /root.acl <- the root ACL
    /root/resource (inherits /root.acl)
    /root/other (inherits /root.acl)

If a request to `/root/resource` returns a link header pointing to the ACL at `/root.acl`, and a client modifies that ACL with the intention of changing the access controls to `/root/resource`, the result would be that, in addition to `/root/resource`, the other two resources (`/root` and `/root/other`) would have had their access controls changed, but the client may be entirely unaware of this change. This could be a _very bad thing_ and lead to the concerns I am alluding to above.

In terms of having a client find the "effective ACL", SOLID has this to say:

> It is considered an anti-pattern for a client to perform those steps, however. A client may discover and load a document's individual resource (for example, for the purposes of editing its permissions). If such a resource does not exist, a client SHOULD NOT search for, or interact with, the inherited ACLs from an upstream container.

birkland:
So with the 404 approach, it's an anti-pattern for the client to try to generate the effective acl.  Does that mean that it could also be an anti-pattern for the client to be _given_ the effective ACL (i.e. that's a piece of information that ought to be internal to the server, clients shouldn't mess with it), or is giving clients the effective acl considered to be a good value-add piece of information to offer?   I ask because I simply don't know the norms and conventions.

As far as the effective acl approach, if a Fedora implementation wanted to give the user a view of the effective acl that they can then modify, it would probably have to present that as follows:

<pre>
/root
/root.acl <- the root ACL
/root/resource
/root/resource.acl (contains the effective acl, as derived/inherited from /root.acl)
/root/other
/root/other.acl ( contains the effective acl, as derived/inherited from /root.acl)
</pre>

so the representation of `/root/resource.acl` and `root/other.acl` could be generated dynamically.  Then if modified (if /`root/resource.acl` is acted upon by `PUT` or `PATCH`), the result is persisted as that resource's proper acl as a regular `LDP-RS`.

.. at least, it seems to me that'd be the way you'd have to do it in order to avoid the problem you're describing @acoburn.  That may be a reasonable choice for an implementation, but sounds fairly burdensome or complicated to _require_.

acoburn:
@birkland if an implementation wants to provide an "effective ACL", there is nothing stopping it from doing this:

    Link: <http://localhost/repository/root.acl>; rel="http://example.com/EffectiveAcl"
    Link: <http://localhost/repository/root/resource.acl>; rel="acl"

Then, if a client understands the semantics of `ex:EffectiveAcl`, then the data is easily retrievable. I think that would be the better approach.

birkland:
Including some sort of "effective acl" link seems a lot more explicit and clear to me as well.

zimeon:
IMO, the WAC spec doesn't make this effective ACL calculation optional -- it is simply the way that the [server MUST find the ACL for a resource by inheritance](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm), and the effective ACL is exposed with `rel="acl"`

acoburn:
@zimeon "effective ACL" is entirely orthogonal to WebAC. It is pure Fedora.

zimeon:
@acoburn - then I guess we are talking at cross-purposes... I thought you were using that as shorthand for the ACL that is there in effect ;-)

acoburn:
@zimeon the current Fedora WebAC module is implemented like this: `rel="acl"` points to the "effective ACL", but that is incorrect, given the SOLID WebACL spec.

zimeon:
@acoburn - I understand that now, but I think the issue at hand is what the Fedora spec should be. I still do not understand why it would be beneficial to allow an implementation to expose a broken ACL link. I am :+1: on this PR.

acoburn:
@zimeon it's not a "broken ACL link". It is a resource that doesn't (yet) exist. As I said before, this PR conflicts with the SOLID reference implementation of WebACL. I would encourage you to think very, very hard about this PR before merging it.

zimeon:
@acoburn - I disagree that there is a conflict with WAC spec (there may be a difference with the solid implementation). Specifically, in your https://github.com/fcrepo/fcrepo-specification/pull/198#issuecomment-324054391 :

> This implies that a document's own ACL resource may not exist. And the implication of that is that such a non-existent ACL document may not be resolvable.

I agree that a particular resource may not have its own ACL. However, the current specification requires that it have an ACL by inheritance (WAC and Fedora) or the default (a Fedora addition). The  location of this ACL which will be exposed in a `rel="acl"` Link. Since one doesn't expose the URI for a non-existent ACL in the case that a particular resource doesn't have its own ACL I don't think your second implication follows.

birkland:
@zimeon I'm going to be involved in the upcoming alignment spec, so am also thinking about this from the perspective of implementing it, albeit via modifying the existing Fedora instance.  

Some of your comments imply that for a resource without its own acl, `rel=acl` would contain a URI of an acl resource of one of the resource's ancestors, as discovered by the algorithm.  Is this correct?  

If so, that means that a client shouldn't try to `PATCH` or `PUT` to it as a means of altering _that particular resource's_ acl, since it could be some other resource's (and the default for thousands of others).

Is this understanding correct so far?

zimeon:
@birkland - yes, the WAC spec is silent on how such an update might occur and the current Fedora spec does not fill the gap either (I think it should)

acoburn:
@zimeon this is simply not correct. A `rel="acl"` link refers to a resource's own ACL resource. It is most certainly _not_ supposed to refer to some ancestor's ACL.

acoburn:
@zimeon while the SOLID WebACL reference implementation is not the only way to implement WebACL, if Fedora claims that it, too, is implementing the SOLID WebACL specification, it most certainly shouldn't add something that outright contradicts the SOLID WebACL reference implementation. Doing so strikes me as seriously wrong.

zimeon:
@acoburn thanks, I think this is the key difference in our understandings. I agree that in your interpretation it would not make sense to disallow ACLs that don't yet exist. I need to spend some time reading the WAC spec again...

birkland:
@zimeon @acoburn  For me at least, I think this is the crux of my misunderstanding.  Is the semantics of the acl link "this is the acl that shall be applied to this resource", or is it "this is the resource's own distinct acl"?

zimeon:
@birkland agreed, that is the question

acoburn:
From the SOLID WebACL specification:

> ACL Resource Location Discovery

> Given a URL for an individual resource or container, a client can discover the location of its corresponding ACL by performing a HEAD (or a GET) request and parsing the rel="acl" link relation.

That is, the `rel="acl"` link points to a resource's corresponding ACL.

In the inheritance section (SOLID WebACL):

> 1. Use the document's own ACL resource if it exists (in which case, stop here).
> 2. Otherwise, look for authorizations to inherit from the ACL of the document's container. If those are found, stop here.

My understanding from this is that a resource's own ACL (that which is pointed to via `rel="acl"`) may not exist (e.g. it may return a 404). And that is exactly how it works in the SOLID reference implementation.

birkland:
I wish `acl` makes it some day into the IANA registry,..

But anyway, I tend to agree with the semantics of `rel=acl` as linking to 'the resource's own distinct (i.e. corresponding) ACL', in which case a 404 would be meaningful.

zimeon:
I (finally) agree with @acoburn and @birkland that the intention of the WAC specification is that `rel="acl"` will always point the individual resource ACL (and not one found by inheritance). I think the WAC spec could be clarified by a change from:

> Given a URL for an individual resource or container, a client can discover the location of its corresponding ACL by performing a `HEAD` (or a `GET`) request and parsing the `rel="acl"` link relation.

to

> Given a URL for an individual resource or container, a client can discover the location of its individual ACL by performing a `HEAD` (or a `GET`) request and parsing the `rel="acl"` link relation.

I also looked at the [node-solid-server](https://github.com/solid/node-solid-server) implementation and confirm that it returns and individual ACL URI in the response to every request by blindly appending `.acl` to the request URI (whether or not the resource or ACL exist).

Given that we require a `rel="acl"` link for every resource, mandating that it resolves (excepting perhaps via redirect to the effective URI -- which I don't think that is a great idea) seem wrong and would short-circuit any possible use of inheritance. As @acoburn has argued, a benefit of having the `rel="acl"` link to 404 provides a mechanism by which a client might know where to PUT an individual resource ACL. I'm thus :-1: on this PR and propose that we close it. (Will discuss alternative clarification in #175...)

escowles:
I agree with @acoburn and @zimeon that the `Link rel="acl"` header should point to the resource's individual ACL, not any inherited ACL, even if the resource does not yet have an individual ACL.  That URI does not need to exist, but it should be able to be created, providing a mechanism for the server to advertise the location a client can PUT an ACL document for the resource.  So I'm 👎  on this PR as well.

zimeon:
Closing as we no longer think this PR is right.

Issue #166, opened 2017-07-14
Requiring WebIDs?


awoods:
The [Solid spec](https://github.com/solid/web-access-control-spec#overview) requires to use of WebIDs.
[We currently support strings and URIs for acl:Agent objects](https://wiki.duraspace.org/display/FEDORA4x/WebAC+Authorization+Delegate#WebACAuthorizationDelegate-Authorizations).

Do we want to change this to WebIDs?

acoburn:
The use of strings for agents is problematic for many reasons. Using IRIs is the right approach here (WebIDs are simply an example -- a good one at that -- for the use of IRIs for referring to agents).

dannylamb:
Is the WebID anything other than an IRI for a profile page?  We have our own in Islandora and don't require one from Fedora per se.

Does this imply that authorization would be done against this IRI which may or may not represent a user account in Fedora?

awoods:
Here is the brief description of [WebID](https://www.w3.org/2005/Incubator/webid/spec/identity/):

> A WebID is an HTTP URI which refers to an Agent (Person, Organization, Group, Device, etc.). A description of the WebID can be found in the Profile Document, a type of web page that any Social Network user is familiar with.
> 
> A WebID Profile Document is a Web resource that must be available as text/turtle [turtle], but may be available in other RDF serialization formats (e.g. [RDFA-CORE]) if requested through content negotiation.
> 
> WebIDs can be used to build a Web of trust using vocabularies such as FOAF [FOAF] by allowing people to link together their profiles in a public or protected manner. Such a web of trust can then be used by a Service to make authorization decisions, by allowing access to resource depending on the properties of an agent, such that he/she is known by some relevant people, works at a given company, is a family member, is part of some group, etc..

I am not sure we can commit to full WebID... but at least require user-URIs.

zimeon:
I think that from a Fedora spec p.o.v., all we need to agree is that there will be agent-URIs that are used to match against these [URIs in WAC](https://github.com/solid/web-access-control-spec#describing-agents). Only in the case of `acl:agentGroup` is there an expectation to be able to resolve that agent-URI to a document describing the group.

Unless there is a compelling reason to continue support strings I would let that go and align with WAC. An implementation could use some unresolvable URI pattern to migrate from strings if necessary.

dannylamb:
:+1: to removing support for strings and expecting URIs

PR #199, opened 2017-08-21
Require IRIs for users and groups

awoods:
Resolves: https://github.com/fcrepo/fcrepo-specification/issues/166

### DIFF
--- a/index.html
+++ b/index.html
@@ -863,7 +863,7 @@
           As a departure from the [[SOLIDWEBAC]]
           <a href="https://github.com/solid/web-access-control-spec#overview">requirement</a>
           that users and groups be identified by WebIDs, this specification requires that users and groups MUST be
-          described by any IRI.
+          identified by any IRI.
         </p>
       </section>
       <section id="link-rel-acl">
--- a/index.html
+++ b/index.html
@@ -857,6 +857,15 @@
           resource SHOULD be located in the same server as the controlled resource.
         </p>
       </section>
+      <section id="agents">
+        <h2>Describing Agents</h2>
+        <p>
+          As a departure from the [[SOLIDWEBAC]]
+          <a href="https://github.com/solid/web-access-control-spec#overview">requirement</a>
+          that users and groups be identified by WebIDs, this specification requires that users and groups MUST be
+          described by any IRI.
+        </p>
+      </section>
       <section id="link-rel-acl">
         <h2>ACLs are discoverable via Link Headers</h2>
         <p>


acoburn:
The SOLID WAC document describes Agents like so:

> In WAC, we use the term Agent to identify who is allowed access to various resources. In general, it is assumed to mean "someone or something that can be referenced with a WebID" ....

I do not understand this to mean that "Agents MUST be described using WebIDs"; rather, I understand this to mean that WebIDs are one valid construct for describing agents. As such, I do not understand the need for the addition in this PR; I find it redundant and I do not agree that it is "a departure from the SOLIDWEBAC requirement".

awoods:
According to the [SOLID overview](https://github.com/solid/web-access-control-spec#overview):
> 1. The resources are identified by URLs, and can refer to any web documents or resources.
> 2. It is declarative -- access control policies live in regular web documents, which can be exported/backed easily, using the same mechanism as you would for backing up the rest of your data.
> 3. Users and groups are also identified by URLs (specifically, by WebIDs)

My reading is that, specifically, WebIDs are required.
This PR is intended to clarify that WebIDs are not required, nor URLs, but rather IRIs.

acoburn:
Given a group IRI: `info:group/foo`, which is a valid IRI (and valid according to this PR), how would an implementation follow the [SOLID guidelines](https://github.com/solid/web-access-control-spec#groups-of-agents) to dereference that group?

acoburn:
Perhaps I should ask: what is the rationale behind departing from the SOLID specification like this? I can support declarations in the Fedora specification that are _more specific_ than in a parent specification (whether that is LDP, HTTP, etc), but stating that a requirement of a parent specification does not apply seems very problematic. It is akin to asserting that the HTTP header `Expect` can be used with 2xx responses. Doing so will weaken the Fedora specification, and I would encourage you to examine exactly why you are choosing to depart from a parent specification.

I would broadly state that I am 👎  on any such departure from a parent specification.

awoods:
Changing this wording from `IRI` to `URL` probably makes sense.

escowles:
Related: I have an action item to add language explictly requiring that group IRIs be resolvable, and reference members using `vcard:hasMember`.  I have added #200 to track that.

acoburn:
@escowles doesn't that mean that Fedora requires that groups are, effectively, WebIDs (they are URLs and they can be dereferenced as RDF resources)? SOLID already requires that groups are identified as URLs and that they can be dereferenced as RDF, and so, why does Fedora need to restate something that SOLID already includes?

acoburn:
@awoods: if WebID is a proposed W3C standard and SOLID embraces it for use with WebAC, why would Fedora explicitly _not_ use WebIDs for agents and groups? Isn't a major goal of the Fedora specification effort designed to align Fedora with other, existing web standards?

awoods:
The intent of this PR is to clarify that although the SOLID document can be interpreted as requiring WebIDs, Fedora makes no such requirement.

acoburn:
@awoods: and why not? I would expect at least some non-normative text explaining that decision.

awoods:
Are you now suggesting that WebIDs be required?

acoburn:
@awoods: I am suggesting that Fedora not depart from SOLID. If that means that Fedora requires using WebIDs, then yes, that is what I am suggesting. This does not mean that I would include text in the Fedora specification that explicitly states that agents MUST be represented by WebIDs, but I would think it would make sense to have some non-normative text referencing the SOLID specification about representing agents/groups and the importance of making those identifiers be URLs.

escowles:
I would be fine with saying to use SOLID, but having a non-normative note saying basically what's in #200 — groups must be dereferencable and point to user URIs with `vcard:hasMember`.  That would be good explicit guidance about how we interpret SOLID and what we think is the minimum an implementation could do to be useful and valid.

escowles:
Closing in favor of #207

PR #201, opened 2017-08-31
Add note about acl:default instead of obsolete acl:defaultForNew

### LINKED ISSUE #164: Requiring acl:defaultForNew?
[acl:defaultForNew](https://github.com/solid/web-access-control-spec#default-inherited-authorizations) - we currently use an assumed inheritance. Is the Fedora spec requiring a change to this other pattern? 

Extracted from: https://github.com/fcrepo/fcrepo-specification/issues/145

zimeon:
Fixes #164

### DIFF
--- a/index.html
+++ b/index.html
@@ -887,6 +887,14 @@
           resource, then the server MUST supply a default <a>ACL</a>. The default <a>ACL</a> resource SHOULD be located
           in the same server as the controlled resource.
         </p>
+        <blockquote class="informative">
+          <p>
+            Non-normative note:
+            The predicate <code>acl:default</code> in the [[WEBAC]] vocabulary indicates authorizations that should be
+            inherited. The [[SOLIDWEBAC]] documentation is being updated to use this in place of the obsolete predicate
+            <code>acl:defaultForNew</code>.
+          </p>
+        </blockquote>
       </section>
     </section>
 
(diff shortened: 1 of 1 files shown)


escowles (dismissed):

index.html
```diff
@@ -887,6 +887,14 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
           resource, then the server MUST supply a default <a>ACL</a>. The default <a>ACL</a> resource SHOULD be located
           in the same server as the controlled resource.
         </p>
+        <blockquote class="informative">
+          <p>
+            Non-normative note:
+            The predicate <code>acl:default</code> in the [[WEBAC]] vocabulary indicates authorizations that should be
+            inherited. The [[SOLIDWEBAC]] documentation is being updated to use this in place of the obsolete predicate
```
acoburn: wouldn't it be better to reference the version of the SOLID spec that "is being updated" to avoid potential confusion after that spec is updated?

zimeon: I'm hoping that this issue will be resolved (and the comment removed) before we finalize the Fedora spec... but I agree that adding the version number might avoid potential confusion.

acoburn: @zimeon that new text is great, thanks so much!

awoods (approved):

escowles (approved):

barmintor (approved):

PR #206, opened 2017-09-06
Adding section clarifying acl:Append

### LINKED ISSUE #170: Require acl:Append and acl:Control
The [Solid spec](https://github.com/solid/web-access-control-spec#modes-of-access) requires the use of 'acl:Append' and 'acl:Control'. 

Do we want to require the same support?

escowles:
Fixes #170

### DIFF
--- a/index.html
+++ b/index.html
@@ -881,8 +881,7 @@
           The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#modes-of-access">Modes of
           Access</a> section mentions the <code>acl:mode</code> value <code>acl:Append</code>.  In the context of a
           Fedora implementation, <code>acl:Append</code> should be understood as operations that only append, such as
-          <code>POST</code>ing to a container, or performing a <code>PUT</code> or <code>PATCH</code> that only adds
-          triples.
+          <code>POST</code>ing to a container, or performing a <code>PATCH</code> that only adds triples.
         </p>
       </section>
       <section id="inheritance">
--- a/index.html
+++ b/index.html
@@ -883,6 +883,38 @@
           Fedora implementation, <code>acl:Append</code> should be understood as operations that only append, such as
           <code>POST</code>ing to a container, or performing a <code>PATCH</code> that only adds triples.
         </p>
+        <section id="appendLDPRS">
+          <h3>LDP-RS</h3>
+          <p>
+            When a client is allowed to perform <code>acl:Append</code> operations on an <a>LDP-RS</a>:
+          </p>
+          <ul>
+            <li>A <code>DELETE</code> request MUST be denied</li>
+            <li>A <code>PATCH</code> request that deletes triples MUST be denied</li>
+            <li>A <code>PATCH</code> request that only adds triples SHOULD be allowed</li>
+            <li>A <code>PUT</code> request on an existing resource MUST be denied</li>
+            <li>A <code>PUT</code> request to create a new resource MUST be allowed if the implementation supports
+              creating resources using <code>PUT</code> (see: <a href='#httpPUTcreate'></a>)</li>
+          </ul>
+        </section>
+        <section id="appendLDPC">
+          <h3>LDPC</h3>
+          <p>
+            In addition to requirements in <a href='#appendLDPRS'></a>, when a client is allowed to perform
+            <code>acl:Append</code> operations on an <a>LDPC</a>, a <code>POST</code> request MUST be allowed.
+          </p>
+        </section>
+        <section id="appendLDPNR">
+          <h3>LDP-NR</h3>
+          <p>
+            When a client is allowed to perform <code>acl:Append</code> operations on an <a>LDP-NR</a>:
+          </p>
+          <ul>
+            <li>All <code>DELETE</code>, <code>POST</code>, and <code>PUT</code> requests MUST be denied</li>
+            <li>A <code>PATCH</code> request that deletes or modifies existing content MUST be denied</li>
+            <li>A <code>PATCH</code> request that only adds content SHOULD be allowed</li>
+          </ul>
+        </section>
       </section>
       <section id="inheritance">
         <h2>Inheritance and Default ACLs</h2>
--- a/index.html
+++ b/index.html
@@ -875,6 +875,16 @@
           Location Discovery</a>.
         </p>
       </section>
+      <section id="append-mode">
+        <h2>Append Mode</h2>
+        <p>
+          The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#modes-of-access">Modes of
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -875,6 +875,16 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
           Location Discovery</a>.
         </p>
       </section>
+      <section id="append-mode">
+        <h2>Append Mode</h2>
+        <p>
+          The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#modes-of-access">Modes of
+          Access</a> section mentions the <code>acl:mode</code> value <code>acl:Append</code>.  In the context of a
+          Fedora implementation, <code>acl:Append</code> should be understood as operations that only append, such as
+          <code>POST</code>ing to a container, or performing a <code>PUT</code> or <code>PATCH</code> that only adds
```
dannylamb: We should also consider PUT to creating resources as well as just POSTing.

acoburn: @dannylamb there is some disagreement about whether create-on-PUT adds `ldp:contains` triples to the parent resource. While Fedora4 allows this behavior, other implementations may not, and after much discussion, the Fedora specification has, thus far, decided to remain silent on this point. @escowles' text above (in my reading) encompasses the possibility that a PUT would add triples to the repository -- whether that relates to the target resource or some other resource such as a parent container.

dannylamb: @acoburn @escowles Fair enough.

dannylamb (dismissed):

zimeon (dismissed):

index.html
awoods: I would be inclined to refine `acl:Append` to only include:
- `POST` to a container, and
- `PATCH` to an LDPRS that is only additive.

Although the effect of a PUT that replaces triples with the original plus additional triples is additive, the process is a full over-write.

escowles: @awoods I agree it's harder to tell if a PUT is only additive or not.  I've removed it from the list of examples.

@dannylamb @zimeon My new commit has dismissed your reviews, so please take a look again.

zimeon (approved):

index.html
```diff
@@ -875,6 +875,15 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
           Location Discovery</a>.
         </p>
       </section>
+      <section id="append-mode">
+        <h2>Append Mode</h2>
+        <p>
+          The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#modes-of-access">Modes of
+          Access</a> section mentions the <code>acl:mode</code> value <code>acl:Append</code>.  In the context of a
+          Fedora implementation, <code>acl:Append</code> should be understood as operations that only append, such as
```
awoods, line 883: I agree with the sentiment of this PR. Would it be helpful to add MUST/SHOULD language so that this text is more directly testable?

dannylamb:
@escowles Bump.  Would you consider adding normative language around this or are you specifically trying to avoid it?

awoods (changes requested):
See: https://github.com/fcrepo/fcrepo-specification/pull/206#pullrequestreview-65620219

index.html
escowles: @awoods I think some testable outcomes of this are, when a client makes a request whose effective ACL allows the authenticated user to `acl:Append`:
* POST MUST be allowed
* PATCH that deletes triples MUST be denied
* PATCH that only adds triples SHOULD be allowed (SHOULD here to allow implementations the option of not checking and just denying all PATCHes)
* PUT MUST be denied
* DELETE MUST be denied

Is there anything missing from this list?

awoods: That sounds perfect, @escowles . Would you be interested in adding the clarity of your bullets in the text of the PR?

escowles: Yes, I'll update this PR with the normative language — just wanted to make sure we agreed on the additions first.

zimeon: @escowles - seems like there are different cases mixed up here depending on the resource type affected:

LDP-NR:
  * POST, PATCH, PUT, DELETE MUST all be rejected

LDP-RS:
  * PATCH that deletes triples MUST be denied
  * PATCH that only adds triples SHOULD be allowed (SHOULD here to allow implementations the option of not checking and just denying all PATCHes)
  * PUT to replace the LDP-RS MUST be denied
  * DELETE MUST be denied

LDPC has LDP-RS rules above plus:
  * POST of a new resource in the LDPC MUST be allowed
  * PUT of a new resource in the LDPC MAY be allowed (per https://fcrepo.github.io/fcrepo-specification/#httpPUTcreate )

zimeon (changes requested):
I agree with @escowles that testable language would be good

index.html
escowles: @zimeon I was just drafting the language and starting to think about the implications of different resource types, so thanks for laying it out so clearly!

birkland: I'd disagree about LDP-NR "POST, PATCH, PUT, DELETE MUST all be rejected".  It's possible that an implementation may be able to implement PATCH in a meaningful way for certain media types, which presumably leads to the possibility of appending to it.  The language in the spec ought to make appending to binaries allowable.

escowles:
@awoods @zimeon @birkland I've just added a block of explicit requirements, broken down by LDP-RS, LDPC, and LDP-NR.

index.html
zimeon: I agree with @birkland's comment and that seems handled by @escowles new changes

zimeon (approved):

awoods (approved):

barmintor (approved):
Nice work here.

index.html
```diff
@@ -875,6 +875,47 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
...
+        <section id="appendLDPRS">
+          <h3>LDP-RS</h3>
+          <p>
+            When a client is allowed to perform <code>acl:Append</code> operations on an <a>LDP-RS</a>:
+          </p>
+          <ul>
+            <li>A <code>DELETE</code> request MUST be denied</li>
+            <li>A <code>PATCH</code> request that deletes triples MUST be denied</li>
+            <li>A <code>PATCH</code> request that only adds triples SHOULD be allowed</li>
+            <li>A <code>PUT</code> request on an existing resource MUST be denied</li>
```
dannylamb, line 895: Maybe this should be a SHOULD?  You could PUT the original resource plus some new triples.

awoods: I agree that the result would appear to be the appending of triples, but the semantics of `PUT` is to remove the existing triples and replace them with the provided triples... which is different than "append".

I would prefer we stick with the existing `MUST` language.

dannylamb (approved):

index.html
dannylamb: I'm fine with that.

PR #207, opened 2017-09-06
Clarifying WebID requirement

### LINKED ISSUE #166: Requiring WebIDs?

### LINKED ISSUE #200: Group URIs must be dereferencable
Related to #166, the editors decided that group URIs must be dereferencable, and must be RDF, and must link to member URIs with `vcard:hasMember`.  These are the minimum requirements for clients and servers to be able to figure out which users are included in a group.

escowles:
Fixes #200, Fixes #166, Supersedes #199

### DIFF
--- a/index.html
+++ b/index.html
@@ -888,6 +888,26 @@
           in the same server as the controlled resource.
         </p>
       </section>
+      <section id="webids">
+        <h2>WebIDs</h2>
+        <blockquote class="informative">
+          <p>
+            Non-normative note:
+            The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#overview">overview</a>
+            mentions parenthetically that user and group identifiers should be WebIDs.  For the purposes of performing
+            authorization, the essential requirements are:
+          </p>
+          <ul>
+            <li>
+              User and group identifiers must be IRIs.
+            </li>
+            <li>
+              Group IRIs must be dereferencable as RDF, and must link to member IRIs using the
+              <code>vcard:hasMember</code> predicate.
+            </li>
+          </ul>
+        </blockquote>
+      </section>
     </section>
 
     <section id="notifications">


index.html
```diff
@@ -888,6 +888,26 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
...
+        <blockquote class="informative">
+          <p>
+            Non-normative note:
+            The [[!SOLIDWEBAC]] <a href="https://github.com/solid/web-access-control-spec#overview">overview</a>
+            mentions parenthetically that user and group identifiers should be WebIDs.  For the purposes of performing
+            authorization, the essential requirements are:
+          </p>
+          <ul>
+            <li>
+              User and group identifiers must be IRIs.
```
acoburn: One key piece about WebIDs is that they are not just IRIs but also URLs. Is that requirement not part of this? Requiring a URL is implied below in that "Group IRIs must be dereferenceable", so would it make sense to state that more explicitly? And I assume from this that User IRIs don't necessarily need to be URLs, correct?

index.html
```diff
@@ -888,6 +888,26 @@ <h2 id='ldpcvpost'>HTTP POST</h2>
...
+            </li>
+            <li>
+              Group IRIs must be dereferencable as RDF, and must link to member IRIs using the
```
acoburn: dereferencable => dereferenceable

acoburn: also, the comma is not necessary.

zimeon (approved):

awoods (approved):

barmintor (approved):
Thanks for this, escowles.

dannylamb (approved):

PR #212, opened 2017-09-08
Add note about non-constraint of RDF data storage model

### LINKED ISSUE #93: Should the specification say something about the connectedness of RDF behind multiple resources?
One of the most striking features of the LPD specification is that it is silent on whether or not LDP-RSs are separate buckets of data (e.g. files, named graphs), or views into a shared dataset (e.g. a triplestore with one graph). Fedora4 effectively sidesteps this distinction through the subject restriction on triples in a LDP-RS: the subject of each triple must be either the resource or a fragment of the resource it is in.

Should the new specification restrict implementations to a particular approach? Should it say anything about this? I raised this issue at the [LDCX17 Standardizing Fedora session](https://docs.google.com/document/d/1m9QwMimrUqFVBH8ZEx79tpYFSczQPgUT8BNemsIk5P0/edit) because it seems to tie with core questions about what sort of interoperability the specification is designed to provide.

I suspect the consensus will be that the specification should not impose restrictions in this area. If that is the case then we should be clear that "swappable" (e.g. as discussed [2017-03-23](https://wiki.duraspace.org/x/3Qv2B)) is not a goal of this specification (though could be combined with others for such purposes), and such understanding might help us think about other issues.

zimeon:
Fixes #93.

Adds text close to that suggested in https://github.com/fcrepo/fcrepo-specification/issues/93#issuecomment-318377226 .

### DIFF
--- a/index.html
+++ b/index.html
@@ -225,6 +225,14 @@
   </section>
     <section id="resource-management">
       <h2>Resource Management</h2>
+      <blockquote class="informative">
+        Non-normative note: Neither [LDP] nor this specification constrains the underlying storage model for RDF
+        data. Each <a>LDP-RS</a> may be entirely independent RDF content, may be a portion of a shared RDF store,
+        or follow some other arrangement. Implementation choices that result in failure to complete client requests
+        are expressed in a constraints document ([[LDP]]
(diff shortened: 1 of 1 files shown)


awoods (approved):

dannylamb (approved):

escowles (approved):

awoods:
@barmintor : this is ready when you are.

PR #213, opened 2017-09-08
Declare the two namespace prefixes we use

zimeon:
We use `ldp:` and `acl:` namespace prefixes without declaring them (or explicitly stating that they are as defined in the LDP and ACL specs). It seems most clear to simply define them.


awoods (approved):

dannylamb (approved):
:+1:

escowles (approved):

awoods:
@barmintor : this is ready when you are.

PR #214, opened 2017-09-08
Add section on ACL linking on resource creation

### LINKED ISSUE #176: ACL creation and linking -- be explicitly silent or specify?
The SOLIDWEBAC specification and the Fedora specification are both currently silent on how or when and ACL may be created and linked to the resource(s) it is designed to control access for. The following cases all seem conforming at present:

  * An implementation aiming to provide easy migration from [Fedora4 WebAC](https://wiki.duraspace.org/x/AJUpB) might require an additional type to indicate an ACL (currently an ACL is a BasicContainer with a specific type, which has children for Authorizations -- this has to change to single resource) and use the `acl:accessControl` property on a resource to indicate its ACL (this predicate _is_ in the [WAC ontology](http://www.w3.org/ns/auth/acl#) but is not mentioned in the SOLIDWEWAC spec) 
  * An implementation has a default ACL that applies to all resources and it never creates nor allows creation or linking of any other ACL (assumes #163)
  * One default ACL as previous bullet that can be edited to be the sole source of fine grained authorization data for many resources (sounds like scaling nightmare!)
  * An implementation automagically creates a new ACL for every resource created (based on some implementation policy, which _might (but need not)_  be based on the idea of inheritance at creation) and thus avoids the need for any inheritance when later determining which ACL applies 
  * An implementation allows creation of ACL documents by creation of an LDPRS that _smells like an ACL_ (I say this because ACLs are not typed), and is linked to a resource based on interpretation of the `acl:accessTo` (has rules determining whether linking happens for one or more resources, what to do if one of these already has an ACL, etc.)
  * An implementation allows linking an LDPRS to a resource as its ACL by some other means outside of Fedora and LDP specs
  * ... more I'm sure ...

IMO, we should _either_ be explicitly silent:

> Non-normative note: Behaviors regarding the creation of ACLs and associating them with the resources they control access for are not defined by \[SOLIDWEBAC] or this specification.

_or_ specify some restrictions.

(I believe that, given the understanding that an ACL is an LDPRS, and the `acl:Control` permission (#170), changes to an ACL can be made with PUT/PATCH as normal under LDP.)

zimeon:
Fixes #176.

Adds a section "ACL linking on resource creation" based on https://github.com/fcrepo/fcrepo-specification/issues/176#issuecomment-319149546 . This places a requirement on servers to recognize the `Link: rel="acl"` header on resource creation, whether or not they support specification of an ACL in this way. If they don't support it, response will always be 4xx. It does not preclude other ways of linking a resource to an ACL.

### DIFF
--- a/index.html
+++ b/index.html
@@ -925,6 +925,18 @@
           </ul>
         </blockquote>
       </section>
+      <section id="link-acl-on-create">
+        <h2>ACL linking on resource creation</h2>
+        <p>
+          A client HTTP <code>POST</code> or <code>PUT</code> request to create a new <a>LDPR</a> MAY include a
+          <code>Link: rel="acl"</code> header referencing an existing <a>LDP-RS</a> to use as the <a>ACL</a>
+          for the new <a>LDPR</a>. The server MUST respond with a 4xx range status code (e.g. 409 Conflict) if it
+          isn't able to create the <a>LDPR</a> with the specified <a>LDP-RS</a> as the <a>ACL</a>. In that
+          response, the restrictions causing the request to fail MUST be described in a resource indicated by a
+          <code>Link: rel="http://www.w3.org/ns/ldp#constrainedBy"</code> response header, following the pattern of
+          [[!LDP]] <a href='https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs'>4.2.1.6</a>.
+        <p>
+      </section>
     </section>
 
     <section id="notifications">
--- a/index.html
+++ b/index.html
@@ -930,11 +930,11 @@
         <p>
           A client HTTP <code>POST</code> or <code>PUT</code> request to create a new <a>LDPR</a> MAY include a
           <code>Link: rel="acl"</code> header referencing an existing <a>LDP-RS</a> to use as the <a>ACL</a>
-          for the new <a>LDPR</a>. The server MUST respond with a 4xx range status code (e.g. 409 Conflict) if it
-          isn't able to create the <a>LDPR</a> with the specified <a>LDP-RS</a> as the <a>ACL</a>. In that
-          response, the restrictions causing the request to fail MUST be described in a resource indicated by a
-          <code>Link: rel="http://www.w3.org/ns/ldp#constrainedBy"</code> response header, following the pattern of
-          [[!LDP]] <a href='https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs'>4.2.1.6</a>.
+          for the new <a>LDPR</a>. The server MUST reject the request and respond with a 4xx or 5xx range status
+          code (e.g. 409 Conflict) if it isn't able to create the <a>LDPR</a> with the specified <a>LDP-RS</a> as
+          the <a>ACL</a>. In that response, the restrictions causing the request to fail MUST be described in a
+          resource indicated by a <code>Link: rel="http://www.w3.org/ns/ldp#constrainedBy"</code> response header,
+          following the pattern of [[!LDP]] <a href='https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs'>4.2.1.6</a>.
         <p>
       </section>
     </section>


birkland:
(Just trying to clarify that I understand) 

If no `Link: rel="acl"` is provided by the client, given our understanding that the semantics of the acl rel [point to the resource's own ACL](https://github.com/fcrepo/fcrepo-specification/pull/198#issuecomment-326126675) and the acl rel MUST always be present, think we can conclude that the server will either create such a resource, or link to a nonexistent resource that the client can discover and `PUT` in order to create a resource's ACL.  So #176 already has an implicit solution, I believe.

So, is this PR is adding additional functionality that optionally allows the client to pre-specify the location of the resource's own individual ACL, by pointing it to a resource that already exists?  If a repository returns a success response code, then would the `rel=acl` of the newly created resource inherently be identical to the one specified in the request?  Is the intent, therefore, to let the server optionally provide a mechanism by which many resources can point to (and therefore share) the same ACL resource, explicitly?

acoburn:
👎  for the reasons outlined here: https://groups.google.com/forum/#!topic/fedora-tech/VLDQOwhhTqY

escowles (dismissed):

awoods (dismissed):

zimeon:
Rebased onto current master to fix merge conflict. No change to addition. Apologies for dismissing your reviews @escowles and @awoods

awoods (approved):

escowles (dismissed):

dannylamb (dismissed):

barmintor:
Worth noting: LDP does not require failing with 4xx (just failing) when discussing interaction model support. Should the 4xx requirement be softened to SHOULD, and the MUST reserved for failure in general?

escowles:
I agree with @barmintor here: we could change the language to "The server MUST fail the request if it isn't able to create..."

awoods:
I agree with @barmintor that as it relates to supporting requested interactions model when creating a resource, the LDP spec language states that ["the server MUST fail the request"](https://www.w3.org/TR/ldp/#ldpc-post-createrdf), without mention of constraints.

However, this PR seems less related to interaction models and more related to the general scenario of ["a server that refuses resource creation"](https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs). Without a `constrainedBy` Link header in the response, it may be difficult for the client to determine why the request failed.

As a side note, it seems that a `constrainedBy` Link header in failed interaction model scenarios would also be helpful to the client.

awoods:
@barmintor / @escowles : I believe we are in agreement with the bulk of this PR, the open question is in how to handle the situation where the client-provided `Link: rel="acl"` is not permitted.
1. The current text says: MUST fail with 4xx status and link to `constrainedBy` document.
1. I believe @barmintor's suggestion is to change the above MUST to SHOULD.
1. I believe @escowles' suggestion is to remove the mention of 4xx, but otherwise leave the text as is.

[My inclination](https://github.com/fcrepo/fcrepo-specification/pull/214#issuecomment-338408358) is to leave the wording as is; however, I recall @barmintor had a compeling characterization of the use of `MUST`, `SHOULD`, and `MAY`. 
@barmintor , if you could link to a definition of your conceptualization of the RFC keywords, that may help inform the usage here.

awoods (changes requested):
Pending @escowles / @barmintor input.

zimeon:
Rereading @barmintor's https://github.com/fcrepo/fcrepo-specification/pull/214#issuecomment-335890685 -- I do not understand what failing means if not 4xx response. What am I missing?

escowles:
@zimeon I believe @barmintor was advocating for saying "The server MUST fail the request..." instead of "The server MUST respond with a 4xx range status code (e.g. 409 Conflict)..."

zimeon:
@escowles I agree with your reading but I don't know how to fail the request _without_ returning 4xx

escowles:
@zimeon I agree — as an implementer, I would definitely read either wording to mean I should not create the resource, and instead return a 4xx error.  So it's more about style and specificity, not what the implementation should actually do.

zimeon:
@escowles - OK, then I disagree with @barmintor's requested change on style grounds: every place else we say fail or reject we also specify a 4xx (or more specific) response, so we should here too.

awoods (dismissed):

barmintor:
@zimeon the reason I referred to the interaction model passage from LDP was that it was a comparable instance of intended resource characteristics included in `Link` request headers. I don't think it matters whether the failure is a 4xx, or a 5xx, etc. I also would probably return a 4xx, but I want to approach all the tightening of LDP requirements with clear intention.

zimeon:
Thanks @barmintor, I hadn't understood that was what you had in mind. I tend to think of 5xx as indicating lower level issues only but I agree that LDP does not specify 4xx in this case. I have updated the PR to permit 4xx or 5xx.

barmintor:
Thanks for your patience, @zimeon!

barmintor (approved):

escowles (approved):

awoods (approved):

dannylamb (approved):

PR #219, opened 2017-09-14
Add missing anchors and move some from h2 to section tag

zimeon:
Noticed missing section links in the versioning section in particular...

### DIFF
--- a/index.html
+++ b/index.html
@@ -222,13 +222,13 @@
           contains a machine-readable listing of <a>URI-M</a>s associated to a given <a>URI-R</a>.
         </dd>
       </dl>
-  </section>
+    </section>
     <section id="resource-management">
       <h2>Resource Management</h2>
 
       <section id="general">
         <h2>General</h2>
-	<section id="ldpnr-ixn-model">
+	        <section id="ldpnr-ixn-model">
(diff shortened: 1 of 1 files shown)


barmintor (dismissed):

index.html
```diff
@@ -222,13 +222,13 @@
...
       </dl>
-  </section>
+    </section>
     <section id="resource-management">
       <h2>Resource Management</h2>
 
       <section id="general">
         <h2>General</h2>
-	<section id="ldpnr-ixn-model">
+	        <section id="ldpnr-ixn-model">
```
dannylamb: Indentation looks off here.

zimeon: Rats, was fixing bad indent and managed to put a tab in...

awoods (approved):
Thanks for this clean-up. There is probably some minor value in having a consistent approach to anchor syntax, such as "all lowercase with dashes between words"...
...but this PR is an improvement.

dannylamb (dismissed):

zimeon:
@awoods - I agree that a consistent policy would probably be good but with this PR I just want to move the baseline up a bit without wholesale changing the existing anchors

awoods (dismissed):

Issue #223, opened 2017-09-14
Add GET and HEAD sections for LDPCv


bbpennel:
I would like to propose that the specification include sections dedicated to describing GET and HEAD requests to LDPCv's, and moving the last sentence of the first paragraph from section 4.3.1.

My current suggested text:

> 4.3.2 HTTP GET
An LDPCv must respond to GET Accept: application/link-format as indicated in [RFC7089] section 5 and specified in [RFC6690] section 7.3.
It should respond with the Allow header as outlined in 4.3.x HTTP OPTIONS.
It must respond with a Content-Type header.
It must respond with Accept-Post if Post requests to the LDPCv are allowed.  If POST is supported but bodies are not accepted, then it must respond with Accept-Post: */*; p=0.0
It must respond with Accept-Patch if Patch requests to the LDPCv are allowed.

> 4.3.3 HTTP HEAD
The HEAD method is identical to GET except that the server must not return a message-body in the response.

zimeon:
I agree with the idea of adding these sections, it seems much more logical to put the things like support for `Accept: application/link-format` in a section on GET than in the preceding [General](https://fcrepo.github.io/fcrepo-specification/#general-3) section.

awoods:
@bbpennel : Please feel free to submit a PR.

PR #224, opened 2017-09-15
Added text for HEAD and GET requests to LDPCv objects

bbpennel:
Addresses issue #223

### DIFF
--- a/index.html
+++ b/index.html
@@ -724,10 +724,7 @@
             time-varying representations of the associated <a>LDPR</a>. An <a>LDPCv</a> is both a <a>TimeMap</a> per
             Memento [[!RFC7089]] and an <a>LDPC</a>. As a <a>TimeMap</a> an <a>LDPCv</a> MUST conform to the
             specification for such resources in [[!RFC7089]]. An implementation MUST indicate <a>TimeMap</a> in the
-            same way it indicates the Container interaction model of the resource via HTTP headers. An <a>LDPCv</a>
-            MUST respond to <code>GET Accept: application/link-format</code> as indicated in [[!RFC7089]]
-            <a href='https://tools.ietf.org/html/rfc7089#section-5'>section 5</a> and specified in [[!RFC6690]]
-            <a href='https://tools.ietf.org/html/rfc6690#section-7.3'>section 7.3</a>.
+            same way it indicates the Container interaction model of the resource via HTTP headers.
           </p>
           <p>
             An implementation MUST NOT allow the creation of an <a>LDPCv</a> that is <a>LDP-contained</a> by its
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -741,6 +738,40 @@
...
         <section>
+          <h2 id='ldpcvget'>HTTP GET</h2>
+          <p>
+            An <a>LDPCv</a> MUST respond to <code>GET Accept: application/link-format</code> as
+            indicated in [[!RFC7089]]
+            <a href='https://tools.ietf.org/html/rfc7089#section-5'>section 5</a> and specified in
+            [[!RFC6690]] <a href='https://tools.ietf.org/html/rfc6690#section-7.3'>section 7.3</a>.
+          </p>
+          <p>
+            An implementation SHOULD respond with the <code>Allow</code> header as outlined in
```
awoods: This should be a `MUST`, no?

index.html
```diff
@@ -741,6 +738,40 @@
...
+            indicated in [[!RFC7089]]
+            <a href='https://tools.ietf.org/html/rfc7089#section-5'>section 5</a> and specified in
+            [[!RFC6690]] <a href='https://tools.ietf.org/html/rfc6690#section-7.3'>section 7.3</a>.
+          </p>
+          <p>
+            An implementation SHOULD respond with the <code>Allow</code> header as outlined in 
+            <a href='#ldpcvoptions'></a>.
+          </p>
+          <p>
+            An implementation MUST respond with a <code>Content-Type</code> header.
```
awoods: It seems like this is potentially crossing a line of over-specifying basic HTTP expectations. Is there a particular that we would want to call out the `Content-Type` header?

index.html
```diff
@@ -741,6 +738,40 @@
...
+          </p>
+          <p>
+            If an <a>LDPCv</a> supports <code>POST</code>, then it MUST respond with
```
awoods: I would be ok removing this redundancy with: 4.3.5 HTTP POST

index.html
```diff
@@ -741,6 +738,40 @@
...
+          <p>
+            An implementation MUST respond with a <code>Content-Type</code> header.
+          </p>
+          <p>
+            If an <a>LDPCv</a> supports <code>POST</code>, then it MUST respond with 
+            <code>Accept-Post</code>. If <code>POST</code> is supported but bodies are not
+            accepted, then the <a>LDPCv</a> MUST respond with <code>Accept-Post: */*; p=0.0</code>.
+          </p>
+          <p>
+            An implementation MUST respond with <code>Accept-Patch</code> if <code>PATCH</code>
```
awoods, line 750: I am not exactly sure what we would expect with a `PATCH` on the TimeMap (LDPCv)... but if we continue to support the verb, I would suggest moving this text down to: 4.3.4 HTTP PATCH

index.html
```diff
@@ -741,6 +738,40 @@
...
+            requests to the <a>LDPCv</a> are allowed.
+          </p>
+        </section>
+        
+        <section>
+          <h2 id='ldpcvhead'>HTTP HEAD</h2>
+          <p>
+            The <code>HEAD</code> method is identical to <code>GET</code> except that the server MUST
```
awoods, line 758: Seems redundant.

awoods (changes requested):

index.html
zimeon: I don't think we should allow client changes to the TimeMap, I think this should be just a server generated thing (so not PUT or PATCH)

awoods:
@bbpennel : When you have a moment, please give the review comments a look. Thanks!

index.html
bbpennel: I can remove this, I had included it since it seemed that an implementer would look to the section about GET or HEAD requests to determine which headers would be included in those responses.  When reviewing the specification I found I would miss components that needed to be included in certain interactions because the details of them were listed in other API request sections.

index.html
bbpennel: I was mirroring the way that HEAD is defined section 3.6.  LDPCvs are fairly distinct from other LDPRs so I figured it would make sense to spell expectations for them in similar detail.  I don't feel strongly about it though, so if it doesn't feel like this adds anything I can go ahead and remove it

index.html
bbpennel: resolved

index.html
bbpennel: resolved

index.html
bbpennel: Since an LDPCv is an LDPC, if we are disallowing requests that are allowed by LDPCs I would think we would want to specify that.  My reasons for including this here are the same as Accept-Post, so I will move them both if that is preferable.

escowles:
Closing in favor of https://github.com/fcrepo/fcrepo-specification/pull/264

Issue #217, opened 2017-09-13
Restoring versions


escowles:
The Modeshape implementation currently restores versions using a PATCH request on the version resource.  But 4.2.6 LDPRm PATCH says "An implementation must not support PATCH for LDPRms."

Is there an alternate way of restoring versions?  Could we change that to "An implementation must not support PATCH for **modifying** LDPRms." (inserting "modifying")?

zimeon:
I don't think the **modifying** approach would make sense without extra wording/specificity around the concept of restoring.

I'm not familiar with the current Modeshape implementation: does restore via patch rely on some secretly saved copy? Or is it a way for a client to (re)write history?

zimeon:
Current modeshape implementation spec:  https://wiki.duraspace.org/display/FEDORA4x/RESTful+HTTP+API+-+Versioning#RESTfulHTTPAPI-Versioning-revertBluePATCHReverttoapreviousversionofanobject

escowles:
Maybe doing a PUT on the LDPRv with a Content-Location header pointing at the version would be a better way, and avoid the issue of PATCH not being allowed on the LDPRm?

bbpennel:
I created a pull request to add language related to restoring a version.  If you have a moment please take a look and let me know if it sounds reasonable.

zimeon:
Although I had previously :+1:'d @escowles' https://github.com/fcrepo/fcrepo-specification/issues/217#issuecomment-329582069, I now think that using`Content-Location` in the request to indicate restore-by-reference would go against the [HTTP definition of `Content-Location`](https://tools.ietf.org/html/rfc7231#section-3.1.4.2). I think we need some other by-reference mechanism for both this and the the upload-by-reference use case (mentioned in #210)

zimeon:
I see the the PR #226 uses `Accept-Datetime` as the header to indicate PUT-to-restore.  As noted in https://github.com/fcrepo/fcrepo-specification/issues/215#issuecomment-329502218 , I lean toward repurposing `Memento-Datetime` in requests for this, rather than reinterpreting `Accept-Datetime` for this non-conneg use case.

escowles:
Yes, RFC 2616 conveniently said "The meaning of the Content-Location header in PUT or POST requests is undefined; servers are free to ignore it in those cases."

But RFC 7231 says:
> An origin server that receives a Content-Location field in a request
   message MUST treat the information as transitory request context
   rather than as metadata to be saved verbatim as part of the
[...]

PR #226, opened 2017-09-15
Added LDPRv PUT section for restoring a previous LDPRm

bbpennel:
Addresses #217

### DIFF
--- a/index.html
+++ b/index.html
@@ -601,35 +601,13 @@
 
         <section id='httpput'>
           <h2>HTTP PUT</h2>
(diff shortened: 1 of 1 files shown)


zimeon (approved):
I think we may need to add a few extra words of explanation but I think this PR gets is a step in the right direction.

escowles (approved):

awoods (approved):

dannylamb (approved):
I think this is a great start, and we can improve on it in follow up PRs. :shipit:

PR #227, opened 2017-09-19
Adding note about using Cache-Control: no-cache with fixity checks

### LINKED ISSUE #218: Create non-normative text suggesting Cache-Control for persistence fixity.
When using `Want-Digest` for persistence fixity, the client wants the server to compute (or otherwise verifiably produce) a checksum.  The client ought to be advised that it `SHOULD` use `Cache-Control: no-cache` so that [no intermediate caches/proxies will return a cached response](https://tools.ietf.org/html/rfc7234#section-5.2.1.4).

escowles:
Fixes #218

### DIFF
--- a/index.html
+++ b/index.html
@@ -446,7 +446,7 @@
           <blockquote class="informative">
             Non-normative note: In the presence of a <code>Want-Digest</code> header, the current <code>Digest</code>
             value should be calculated to enable fixity verification. Otherwise, it is expected that ETag values may be
-            cached. Clients should use the <code>Cache-Control: no-cache</code> header (see [[!RFC7234]]
+            cached. Clients should use the <code>Cache-Control: no-cache</code> header (see [[RFC7234]]
             <a href='https://tools.ietf.org/html/rfc7234#section-5.2.1.4'>section 5.2.1.4</a>) to instruct any
             intermediate caches or proxies not to return a cached response. See <a href="#persistence-fixity"></a>.
           </blockquote>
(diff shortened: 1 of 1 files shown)


zimeon (changes requested):
Looks good except `[[!RFC7234]]` should not have the `!` as this is informative (not that that actually has any effect because we do ref this RFC in normative sections...)

zimeon (approved):

barmintor (approved):

dannylamb (approved):

index.html
```diff
@@ -446,7 +446,9 @@
           <blockquote class="informative">
             Non-normative note: In the presence of a <code>Want-Digest</code> header, the current <code>Digest</code>
             value should be calculated to enable fixity verification. Otherwise, it is expected that ETag values may be
-            cached. See <a href="#persistence-fixity"></a>.
+            cached. Clients should use the <code>Cache-Control: no-cache</code> header (see [[RFC7234]]
```
awoods, line 449: Would it be safer to replace `Clients should use` to `Clients are advised to use`?

zimeon: I'm not sure about this because it is the designers of clients that are advised

awoods: To be clear, are you suggesting that the current wording expresses the intent to your liking, @zimeon ?

zimeon: Yes, I think the current wording is adequate (and I think it would not be quite right to say "clients are advised")

PR #232, opened 2017-09-22
Clarifying optionality of server-managed versioning for LDPRv

### LINKED ISSUE #231: Server-managed versioning required?
4.1.3.1 Versioned Resources / PUT / General says:
> An LDPRv may support PUT. An implementation receiving a PUT request for an LDPRv must both correctly respond as per [LDP] as well as create a new LDPRm contained in an appropriate LDPCv.

That sounds like we are requiring server-managed versioning when an LDPR has versions and receives a PUT request.  Should we change that to:
> An LDPRv may support PUT. An implementation receiving a PUT request for an LDPRv must correctly respond as per [LDP] and may also create a new LDPRm contained in an appropriate LDPCv.

escowles:
Fixes #231

### DIFF
--- a/index.html
+++ b/index.html
@@ -604,7 +604,8 @@
           <p>
             An implementation receiving a <code>PATCH</code> request for an <a>LDPRv</a> MUST correctly respond as per
             [[!LDP]] and MAY also create a new <a>LDPRm</a> contained in an appropriate <a>LDPCv</a>. Any newly-created
-            <a>LDPRm</a> SHOULD be the version of the <a>LDPRv</a> that was modified by the <code>PATCH</code> request.
+            <a>LDPRm</a> SHOULD be the version of the <a>LDPRv</a> following modification by the <code>PATCH</code>
+            request.
           </p>
         </section>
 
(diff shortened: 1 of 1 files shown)


zimeon (changes requested):
I think PATCH is currently inconsistent - suggests that the LDPRm is be version _before_ patch, whereas for POST/PUT we are suggesting the version _after_ the request (suggest "... the version of the LDPRv after modification by the PATCH request.").

I'm not really sure that this is the best strategy but we should be consistent, and it is only SHOULD

escowles:
@zimeon I struggled a little with the exact wording, perhaps saying "...as modified by the [method]" would be clearer than "...that was modified by the [method]"?

zimeon:
I think "as modified by" is better than "that was modified by" but still seems ambiguous to me. To make it really clear I can't think of options other than the form "... (after|resulting from|following|...) modification by"

escowles:
@zimeon Yes, that's clearer — I've pushed a new commit that explicitly spells out that it's the state following the creation/modification.

zimeon (dismissed):
Thanks @escowles

awoods (dismissed):

dannylamb (dismissed):
LGTM

Issue #237, opened 2017-09-27
Require support for external content with access-type=URL


escowles:
* Strengthen requirement for external content to MUST, and also require support for access-type=URL
* Strengthen requirement for Want-Digest, etc. for external content to a MUST
* Non-normative note that an implementation may proxy, or copy, or otherwise make permanent external content

barmintor:
If `access-type=URL` is a redirect, does the resulting LDP-NR need to support `Want-Digest`, or does the resource identified in the URL parameter need to support `Want-Digest`?

EDIT: that is to say, if I support this by returning the location in the request's `URL=` parameter as the subsequent response's `Location` header, which resource ought to provide these headers (`Digest`, `Content-Length`, etc)?

awoods:
If you are proxying the content, then Fedora should be providing the `Digest` and `Content-Length` headers.

If you are "redirecting", I think it is fair to assume that Fedora can not guarantee that the external server will support the required headers and therefore, I would suggest that the required headers be the responsibility of the Fedora endpoint, not the external endpoint.

escowles:
I would say in both cases that the Fedora response needs to include those headers.  If the original server provides them, I think they could be copied (that is, I don't think it's Fedora's responsibility to re-calculate the checksums).

PR #242, opened 2017-10-11
Require support for external content with access-type=url

awoods:
* Strengthen requirement for external content to MUST
* Require support for access-type=URL
* Strengthen requirement for Want-Digest for external content to a MUST
* Non-normative note that an implementation may proxy or copy external content

Resolves: https://github.com/fcrepo/fcrepo-specification/issues/237

### DIFF
--- a/index.html
+++ b/index.html
@@ -523,7 +523,7 @@
           and rejection specified here.
         </blockquote>
         <p id='message-external-body-accept-post'>
-          Fedora servers MUST support the creation of LDP-NRs with <code>Content-Type</code> of
+          Fedora servers SHOULD support the creation of LDP-NRs with <code>Content-Type</code> of
           <code>message/external-body</code> and <code>access-type</code> parameter of <code>url</code>.
         </p>
         <p id='message-external-body-supported-type'>
--- a/index.html
+++ b/index.html
@@ -516,16 +516,19 @@
           Non-normative note: Variability among client types and locations may mean that LDP-NR content is addressed in
           ways that are external to the Fedora server but not resolvable by all clients. This specification describes
           the use of <code>Content-Type: message/external-body</code> values to signal, on POST or PUT, that the Fedora
-          server should should not consider the request entity to be the LDP-NR's content, but that a
+          server should not consider the request entity to be the LDP-NR's content, but that a
           <code>Content-Type</code> value will signal a name or address at which the content might be retrieved. The
           <code>url</code> and <code>local-file</code> <code>access-type</code> values motivate this specification, but
           a Fedora server may support any <code>access-type</code> parameters per the requirements for advertisement
           and rejection specified here.
         </blockquote>
         <p id='message-external-body-accept-post'>
-          Fedora servers that support LDP-NR with <code>message/external-body</code> MUST advertise that support in the
-          <code>Accept-Post</code> response header for each supported <code>access-type</code> parameter of supported
-          <code>Content-Type</code> values.
+          Fedora servers MUST support the creation of LDP-NRs with <code>Content-Type</code> of
+          <code>message/external-body</code> and <code>access-type</code> parameter of <code>url</code>.
+        </p>
+        <p id='message-external-body-supported-type'>
+          Fedora servers MUST advertise support in the <code>Accept-Post</code> response header for each supported
+          <code>access-type</code> parameter of <code>Content-Type: message/external-body</code>.
         </p>
         <p id='message-external-body-unsupported-type'>
           Fedora servers receiving requests that would create or update an LDP-NR with a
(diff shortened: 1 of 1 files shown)


escowles (dismissed):

zimeon (dismissed):

index.html
```diff
@@ -516,16 +516,19 @@
...
           <code>Content-Type</code> value will signal a name or address at which the content might be retrieved. The
           <code>url</code> and <code>local-file</code> <code>access-type</code> values motivate this specification, but
           a Fedora server may support any <code>access-type</code> parameters per the requirements for advertisement
           and rejection specified here.
         </blockquote>
         <p id='message-external-body-accept-post'>
-          Fedora servers that support LDP-NR with <code>message/external-body</code> MUST advertise that support in the
-          <code>Accept-Post</code> response header for each supported <code>access-type</code> parameter of supported
-          <code>Content-Type</code> values.
+          Fedora servers MUST support the creation of LDP-NRs with <code>Content-Type</code> of
```
barmintor: I agree with all of the changes here except the addition of this MUST. The other changes are good, even necessary, clarification about what must be done of external content is supported, but the use case is not universal and its support is advertise-able.

barmintor (changes requested):

index.html
awoods: @fcrepo/editors : this specific topic of tightening the language around external content dates back to:
https://wiki.duraspace.org/display/FF/2017-09-27+-+API+Spec+Meeting

The conversations in meetings since 9/27 have included requiring the support for external content. I would like to get @fcrepo/editors input on this suggestion of relaxing that requirement before taking action.

awoods: @fcrepo/editors : Looking for input.

escowles: I don't have a strong opinion on this point.  I would be fine with relaxing that to a SHOULD, for instance.

zimeon: I'm still quite conflicted about what we set as a minimum bar to be "Fedora". In the abstract I agree completely with @barmintor that a client or test-suite can tell whether this feature is supported (I'm imaging a test suite that outputs a feature implementation checklist...) so in that sense SHOULD makes perfect sense. A community agreed set of must-meet vs. optional use cases would really be the right way to arbitrate this but since we don't have that perhaps it is best to go with optional, relax to SHOULD.

escowles: I completely agree with @zimeon here — it would be great to have more solid info about external content.  I know many Samvera repos use it.  But I'm not sure if we need to go so far as to say that all Fedora implementations must support it.

Related to this, I thought we'd coalesced around requiring external content be proxied.  But the discussion in #237 includes discussion of redirecting as an alternative.  The main reason I bring this up is that redirecting seems like a lower bar than proxying, so it might make the difference for me.

awoods: I believe the consensus is that Fedora impls `MUST` support external content, but only `SHOULD` support API creation of such content.
I am happy to change `MUST` to `SHOULD` on [line:526](https://github.com/fcrepo/fcrepo-specification/pull/242/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR526).

index.html
```diff
@@ -553,6 +561,10 @@
           <a href="https://www.w3.org/TR/ldp/#ldprs-put-servermanagedprops">4.2.4.3</a>, 4xx responses MUST be
           accompanied by a <code>Link</code> header with <code>http://www.w3.org/ns/ldp#constrainedBy</code> relation.
         </p>
+        <p id='external-content-want-digest'>
+          <code>GET</code> and <code>HEAD</code> requests to any external binary content <a>LDP-NR</a> MUST correctly
+          respond to the <code>Want-Digest</code> header defined in [[!RFC3230]].
```
escowles, line 566: I like requiring this, and think it will make it more usable for clients.  However, if common webservers (e.g., Apache and Nginx) don't support Want-Digest/Digest, an implementation would need to fetch the content and checksum it in order to generate these headers.  So that would mean fetching the entire content twice to serve it the first time.  Is this acceptable overhead?

index.html
awoods: @barmintor : I believe your requested change has come true.
https://github.com/fcrepo/fcrepo-specification/pull/242/commits/de6233a9a7f9f70d07eaac4f21a1e702c780c9c4

index.html
zimeon: I think this is acceptable overhead now that the requirement for support has been changed to SHOULD

zimeon (approved):

index.html
barmintor: @escowles as implementation work goes forward, we should be trying to identify patterns here. Some of the use cases that motivate external storage are also going to make pre-emptive checksumming especially punitive.

barmintor (approved):

Issue #243, opened 2017-10-11
Remove PATCH on LDPCv


awoods:
An LDPCv (TimeMap) should not support PATCH.

zimeon:
See discussion on PR #244 about whether it is desirable to allow PATCH on LDPCv

PR #244, opened 2017-10-11
Prohibit PATCH on LDPCv

awoods:
Resolves: https://github.com/fcrepo/fcrepo-specification/issues/243

### DIFF
--- a/index.html
+++ b/index.html
@@ -757,8 +757,7 @@
             <a href='#ldpcvdelete'></a> for requirements on <code>DELETE</code> if supported.
           </p>
           <p>
-            An implementation MAY <code>Allow: PATCH</code> if the <a>LDPCv</a> has mutable properties. See
-            <a href='#ldpcvpatch'></a> for requirements on <code>PATCH</code> if supported.
+            An implementation MUST NOT <code>Allow: PATCH</code>.
           </p>
           <p>
             An implementation MAY <code>Allow: POST</code> if versions can be explicitly minted by a client. See
@@ -769,8 +768,7 @@
         <section id="ldpcvpatch">
           <h2>HTTP PATCH</h2>
           <p>
-            An implementation MAY <code>Allow: PATCH</code>, but if so, it MUST NOT permit clients to modify containment
-            triples.
+            An implementation MUST NOT <code>Allow: PATCH</code>.
           </p>
         </section>
 
--- a/index.html
+++ b/index.html
@@ -310,8 +310,8 @@
         </p>
         <section id="httpPATCH-containment-triples">
           <h3>Containment Triples</h3>
-          The server SHOULD NOT allow HTTP PATCH to update an <a>LDPC</a>’s containment triples; if the server receives
-          such a request, it SHOULD respond with a 409 (Conflict) status code.
+          The server SHOULD NOT allow HTTP PATCH to update an <a>LDPC</a>’s containment triples; if the server
+          receives such a request, it SHOULD respond with a 409 (Conflict) status code.
         </section>
         <section id="httpPATCH-ixn-models">
           <h3>Interaction models</h3>
--- a/index.html
+++ b/index.html
@@ -308,6 +308,10 @@
           request MUST respond with a 2xx status code; the specific code in the 2xx range MAY vary according to the
           response body or request state.
         </p>
+        <section id="httpPATCH-containment-triples">
+          <h3>Containment Triples</h3>
+          The server SHOULD NOT allow HTTP PATCH to update an <a>LDPC</a>’s containment triples; if the server receives such a request, it SHOULD respond with a 409 (Conflict) status code.
+        </section>
         <section id="httpPATCH-ixn-models">
           <h3>Interaction models</h3>
           <p>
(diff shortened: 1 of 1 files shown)


birkland:
What is the rationale or underlying principle for prohibiting PATCH on LDPCv?

awoods:
@birkland : I believe the idea is that TimeMaps serve the purpose of containing and listing mementos, period.

birkland:
I think that's too restrictive to set in stone.  Implementations have all the tools they need to lock it down anyway.  For example, is it not allowable _by definition_ to want to call the LDPCv an `ore:Aggregation' (or some other concept that makes sense, according to some ontology)?  Likewise, is it inherently not allowable to do that retroactively via PATCH?

It may be best to just not mention anything (i.e. remove the sections on PATCH).

awoods:
@birkland : I would be fine with dropping the PATCH mentions for LDPCv-s.
@fcrepo/editors : thoughts?

escowles:
Since LDPCv is still a LDPC, it seems good to preserve as much functionality as possible — so I'm fine with leaving those sections as-is.

zimeon:
I'm also OK with leaving PATCH description as-is. Propose closing this PR and #243

awoods:
@birkland @escowles @zimeon : It seems that we are discussing three options as it relates to PATCH on LDPCv-s:
1. Prohibit PATCH (this PR)
2. Be silent on PATCH ( @birkland suggestion )
3. Leave the current text ( @escowles suggestion )

I am inclined towards option-2 as it is not clear to me under what circumstances the user would see user-triples on a TimeMap, given the strict [specification of the contents of the TimeMap](https://tools.ietf.org/html/rfc7089#page-36), but could live with option-3 if that is the consensus view.

escowles:
@awoods We define LDPCv as "A version container: an LDPC that is simultaneously a Memento TimeMap." — since it is an LDPC, then I would expect all of the normal LPDC functionality to work normally.

Memento [section 5](https://tools.ietf.org/html/rfc7089#page-36) says the TimeMap:
* MUST list the URI-R/Original Resource, and the URI-M for each Memento, and unambiguously identify the types of all resources.
* MUST be serializable in the Memento link-value format, but MAY be serializable in other formats.
* SHOULD also include the URI-G/TimeGate and URI-T/TimeMap.

So, I think an RDF description of the LDPRv, LDPCv, and all LDPRms in various RDF formats would fulfill those requirements, and including any other triples that the user had supplied doesn't seem like it would run afoul of the requirements either.

In short, I think the LDPCv can just be a normal container, with some added Memento functionality.  Is there any requirement from LDP or Memento that we suppress any functionality?  Is there any use case for doing so?

awoods:
@escowles : I was thinking mostly about the link-format requirement for TimeMaps:

> The entity-body of a response from a TimeMap MAY be serialized in
   various ways, but the link-value format serialization described here
   MUST be supported.  In this serialization, the entity-body MUST be
   formatted in the same way as the value of an HTTP "Link" header, and
   hence MUST comply to the "link-value" construction rule of Section 5.
   The Link header field of [RFC5988], and the media type of the entity-
   body MUST be "application/link-format" as introduced in [RFC6690].

..and the implication that requirement brings to the content of the TimeMap. However, I could get behind either option-2 or option-3.

zimeon:
@awoods I think it is reasonable that the link format of the TimeMap have only the Memento parts, but that RDF forms might include other triples. I lean slightly towards option-3 (I fear silence leaves obvious questions unanswered) but could also live with option-2.

escowles:
I agree with @zimeon that the link-format of the TimeMap should just have the Memento parts, but the RDF serializations could have other triples.  That's why I think we should keep the current text, since it calls out the possibility of allowing properties on the TimeMap, and identifies the pitfall of allowing modification of the containment triples, which implementations need to avoid.

barmintor:
Should this instead be: "an implementation MUST NOT allow PATCH to modify a LDPCv's containment triples"? It's ambiguous in the LDP spec, but especially problematic in this context. The issue if user triples I prefer to be silent on.

awoods:
Given the diversity of options and opinions here, I have added the topic to discuss during the next API Spec meeting: https://wiki.duraspace.org/display/FF/2017-10-25+-+API+Spec+Meeting

barmintor:
It's been pointed out to me that my suggestion is already in the spec for LDPCv's at 4.3.4 in the 20 Oct draft.

index.html
```diff
@@ -766,14 +770,6 @@
           </p>
         </section>
 
-        <section id="ldpcvpatch">
```
zimeon, line 769: The anchor `#ldpcvpatch` is referenced so that needs to be addressed to allow removal. Otherwise I'm :+1: on the change.

awoods: Updated with: https://github.com/fcrepo/fcrepo-specification/pull/244/commits/cc00a85643e6257d418789c2d9eefdf50361381e

zimeon (dismissed):

escowles (dismissed):

escowles (approved):

zimeon (approved):

zimeon:
Majority approval, no dissenting, was 3 days old before syntax fix -- will merge

PR #257, opened 2017-11-02
Tidy LDPCv section, add "original" to self links

zimeon:
(apologies for going straight to PR but that seems easier than trying to describe changes in abstract)

Two things:

  1. In discussion with @hvdsomp today he pointed out that we say an LDPRv should include header `Link: rel="timegate"` when it should really be `Link: rel="original timegate"` because this is the original an timegate.
  2. @hvdsomp also noticed that we discuss the `Link: rel="original timegate"` header in the [general intro](https://fcrepo.github.io/fcrepo-specification/#LDPRv-general), and then the `Link: rel="timemap"` header under [GET](https://fcrepo.github.io/fcrepo-specification/#LDPRv-get) ... which seems potentially confusing. I took this second point as a cue to try to tidy the LDPRv section a little. Except for point 1, I have not attempted to change meaning although I did add MUST for the inclusion of these headers -- I think they were intended to be required and this is clearly easily testable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -609,9 +609,8 @@
           <h2>General</h2>
           <p>
             A versioned resource (<a>LDPRv</a>) provides a <a>TimeGate</a> interaction model as detailed in the
-            Memento specification [[RFC7089]] and indicated by an HTTP header <code>Link: rel="timegate"</code>
-            referencing itself. It otherwise follows the relevant [[!LDP]] specification with the additional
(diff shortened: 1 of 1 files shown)


awoods (approved):

dannylamb (approved):
I think this is much clearer language.

escowles (approved):

PR #262, opened 2017-11-22
Torching Depth header in lieu of advertising recursive DELETEs 

### LINKED ISSUE #234: Proposal to remove Depth header
The current editor's draft requires some support of the Depth header from RFC4918. This is to provide clients some mechanism for knowing what will be deleted (or attempted to be deleted) when issuing an HTTP DELETE to a resource.

Subsequent conversation on the list and among the editors has suggested that this approach may be specifying too much (requiring contextual support) or not enough (implementation variation may still be unclear and difficult to advertise).

Proposals to address this include: Additional headers (`Accept-Depth`), requiring certain `Depth` values to be supported.

An alternative is to remove mention of `Depth` entirely. This could involve:
1. An informative section indicating that DELETE is understood to be recursive along LDP containment relationships.
2. A requirement that implementations that cannot recurse not advertise DELETE in the OPTIONS response of containers with contained resources.

This approach would:
1. Advertise ability via OPTIONS
2. Clarify client expectations
3. Be testable in the simple case
4. Allow implementations to leverage existing status codes for failure (403, 405, 409), asynchrony (202), or multiple/variable results (207)
5. avoids entanglement with the WebDAV spec, which is complex

A related PR for discussion is forthcoming.

awoods:
Resolves #234

This PR's intention is to remove the Depth header, which has proven confusing, and to replace it by:
* Making sure DELETEs are recursive along the lines of containment
* Allowing implementations that do not want to provide recursive DELETEs by forbidding advertising DELETE in OPTIONS responses for containers with children.

### DIFF
--- a/index.html
+++ b/index.html
@@ -497,7 +497,14 @@
             resources.
           </p>
           <p>
-            An implementation that cannot recurse MUST NOT advertise <code>DELETE</code> in response to OPTIONS
+            An implementation MUST NOT return a 200 or 204 response unless the entire operation successfully completed.
+          </p>
+          <p>
+            An implementation MUST NOT emit a message that implies the successful DELETE of a resource until the
+            resource has been successfully removed.
+          </p>
+          <p>
+            An implementation that cannot recurse SHOULD NOT advertise <code>DELETE</code> in response to OPTIONS
             requests for containers with contained resources.
           </p>
           <blockquote id="httpDELETEatomicity" class="informative">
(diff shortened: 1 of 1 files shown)


escowles (approved):

awoods:
Supersedes: https://github.com/fcrepo/fcrepo-specification/pull/241

dannylamb (approved):
:+1: looks good to me.  thanks for picking this one up @awoods

Issue #233, opened 2017-09-22
Identifying LDPCvs


bbpennel:
Currently LDPRvs and LDPRms are identifiable by their headers, in the first case by the VersionedResource link header, and in the second case by memento headers.  It does not seem like there is currently a way for clients to identify an LDPCv as different from any other LDPC, except perhaps indirectly from its `Accept` header including `application/link-format`.

Given that LDPCvs have a number of interactions that are unique to them from other LDPRs, I wonder if it would make sense to require they respond with an interaction model or other Link header, in a similar way to `http://fedora.info/definitions/fcrepo#VersionedResource`?

birkland:
This would be helpful for messaging too.  If a client receives a message, should/would there be an rdf type that distinguishes it as an LDPCv?

bbpennel:
@birkland I don't think LDPRms would currently be identifiable from non-versionable LDPRs based on RDF types, although being able to do so when filtering messages on rdf type would be very helpful.

bseeger:
@bbpennel  - just a point of clarification -  I don't think that LDPRv's have to return 
a link header containing `http://fedora.info/definitions/fcrepo#VersionedResource`.  They can and it's pretty useful, but I don't think one needs to via the spec.  Via the spec, the only indication that it's a LDPRv is that there's a timegate / timemap link in the headers.

awoods:
Providing `Link: some-type; rel="type"` for the Memento-related resources (timemap, timegate, memento) seem like helpful additions.
...with the following types: http://mementoweb.org/ns.rdf

zimeon:
I agree with @awoods above, adding type `http://mementoweb.org/ns#TimeMap` for the LDPCv would seem very clear.

PR #263, opened 2017-11-22
Datetime negotiation should redirect to the LDPRm

### LINKED ISSUE #258: Should we say datetime negotiation on TimeGate MUST use 302 not 200?
In discussion with @hvdsomp today we noted that the Memento spec is quite permissive on how one gets a Memento from datetime negotiation on a TimeGate. I think in the Fedora/Linked-Data world we would not want to have a 200 response with a Memento, I think we should thus be explicit and say that a 302 redirect is expected (would go in https://fcrepo.github.io/fcrepo-specification/#LDPRv-get-request).

escowles:
* Explicitly require 302 (Found) redirection upon successful Datetime negotiation
* Flatten overly-nested section

Fixes #258

### DIFF
--- a/index.html
+++ b/index.html
@@ -616,28 +616,22 @@
 
         <section id="LDPRv-get">
           <h2>HTTP GET</h2>
-          <section id="LDPRv-get-request">
-            <h2>Request Headers for an LDPRv</h2>
-            <p>
-              The <code>Accept-Datetime</code> header is used to request a past state, exactly as per [[!RFC7089]]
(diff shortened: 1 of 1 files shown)


zimeon (dismissed):

escowles:
Also flatten the 4th-level indentation and combine these two sections.

awoods (approved):

zimeon (approved):

dannylamb (approved):
Approved, but this'll need to get rebased after #233 lands

PR #264, opened 2017-11-22
Clarify LDPCv section

### LINKED ISSUE #223: Add GET and HEAD sections for LDPCv

zimeon:
Suggested replacement for #224 (Added text for HEAD and GET requests to LDPCv objects). Decision not to repeat HEAD sections everywhere so delete from LDPRv and LDPRm also.

Closes #223

### DIFF
--- a/index.html
+++ b/index.html
@@ -650,13 +650,6 @@
             <a href='https://tools.ietf.org/html/rfc7089#section-2.1.1'>section 2.1.1</a>.
           </p>
         </section>
-
-        <section id="LDPRv-head">
-          <h2>HTTP HEAD</h2>
-          <p>
-            See <a href="#httpHEAD"></a> and <a href="#LDPRv-get"></a>.
-          </p>
-        </section>
       </section>
 
       <section id="LDPRm">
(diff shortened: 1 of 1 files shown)


escowles (dismissed):

index.html
```diff
@@ -759,6 +742,21 @@ <h2>General</h2>
...
 
+        <section id='ldpcvget'>
+          <h2>HTTP GET</h2>
+          <p>
+            An <a>LDPCv</a> MUST respond to <code>GET Accept: application/link-format</code> as indicated in
+            [[!RFC7089]] <a href='https://tools.ietf.org/html/rfc7089#section-5'>section 5</a> and specified in
+            [[!RFC6690]] <a href='https://tools.ietf.org/html/rfc6690#section-7.3'>section 7.3</a>.
+          </p>
+          <p>
+            An implementation MUST include the <code>Allow</code> header as outlined in
```
awoods, line 753: This is only a style preference, but these three sentences would read better for me if they were each in their own '\<p\>' block.

awoods (changes requested):
Pending response to one minor comment... if not an actual change.

escowles (approved):

awoods (approved):

dannylamb (approved):

Issue #236, opened 2017-09-27
Use Location header for ingest-by-reference


escowles:
The Location header seems to be defined only for responses (https://tools.ietf.org/html/rfc7231#section-7.1.2), so we can use it for requests to indicate that the server should retrieve the URL and ingest it.

Separating this out from the external-content proxying use case should reduce confusion.

zimeon:
I agree that we should clearly separate out the ingest-by-reference from the proxy cases. So far I think `Location` looks good, but if it turns out no to be then we should probably invest a new header for ingest-by-reference.

barmintor:
Here's my worry here:
1. an implementation doesn't have to support this behavior, but
2. there's no such thing as `Accept-Location`

Should this be an implementation issue instead of a spec issue?

barmintor:
Here are things that I think have consensus:
1. LDP-NR HEAD/GET response includes `Content-Location: foo` means the content is proxied
2. LDP-NR HEAD/GET response includes `Location: foo` means the content is redirected
3. LDP-NR HEAD/GET response includes neither `Content-Location` or `Location` means it is local as of the request
4. `Content-Type: external-body` is the only mechanism we've identified in HTTP 1.1 that allows the server to look elsewhere for the entity
5. the `expiration` parameter on `external-body` may be ambiguous as to whether copying or simply temporality of access is indicated
6. the `Expires` header indicates cache lifespan, and thus cannot be used to indicate actual end of availability

Given the above, would it be clearer to say the `expiration` parameter `SHOULD` result in copied content or an error (as an unsupported `Content-Type1), and advise with recourse to 1-3 above how a client should be able to identify the state of proxying/redirection/hosting in an informative note? As much as I would like to support temporary Location values in Cavendish with an automatic 410, I can't see a cache-friendly way to do that, so maybe the situation is less ambiguous than I've tried to claim.

escowles:
I agree with the list of items in @barmintor 's [comment above](https://github.com/fcrepo/fcrepo-specification/issues/236#issuecomment-338410768).  I think that leaves us with two options:

1. Use `Content-type: message/external-body` with an `expires` parameter to indicate that content should be copied into the repository.  This overloads the mechanism for creating a reference to external content.
2. Use `Location` to indicate remote content that should be ingested.  This separates the two tasks, which I think might reduce confusion.  But it uses an undefined request header, which as we've seen with HTTP 1.1, might be defined to preclude our usage at any time.

[...]

Issue #210, opened 2017-09-06
Clarify intent of external content expiration parameter


birkland:
The current specification draft says that the server SHOULD respect the expiration parameter by copying content.  The external content is explicitly ephemeral in this case, and may disappear.  In general, what sort of guarantee does the repository provide when it returns a successful response to such a request?  Is it the intent of the specification that the repository must make this content durable, and therefore accessible after its expiration date?  Would it be reasonable (and consistent with the intent of the spec) for a repository to remove the corresponding LDP-NR (so that it produces a 410) once the content expires, as an alternative to the copying that it SHOULD do?

zimeon:
ref: last paragraph of https://fcrepo.github.io/fcrepo-specification/#external-content

escowles:
@birkland I think there are two different things going on here:
1. Registering external content to be proxied (where the expiration parameter might just be guidance about how long the content will exist)
1. Ingesting external content by giving the server the URL and expecting it to be retrieved and preserved (where the expiration parameter is used to trigger this behavior).  I'm not sure if this is a good use of the expiration parameter — maybe this should be handled some other way?

birkland:
@escowles I agree with the two different possibilities, and would add that there really isn't any mechanism for the client to express its intent on what strategy the server employs (proxying, copying, redirects, etc).  So the client doesn't know whether the server picks (1) or (2), (or something else) until it happens.  And if the server decides not to include `Content-Location`, then the client might not ever know.   

So the options would be re-cast as:

1. The client provides expiring external content.  The server decides to proxy it, and use the provided expiration data as guidance about how long it will exist.
2.  The client provides expiring external content.  The server decides to retrieve and preserve it.

I believe (2) is a confusing and unintuitive use of the expires header... though acceptable if worded clearly and unambiguously.  The problem is that it's currently ambiguous what the repository will end up doing.

dannylamb:
My reading of that paragraph leads me to believe 2 is the only interpretation because it says the content must be copied.

> Fedora servers receiving requests that would create or update an LDP-NR with a message/external-body content type SHOULD respect the expiration parameter, if present, by copying content.

I'd remove saying the content is copied so that really leaves us with scenario 1, which I think is more in line with the intent of that section.  And it doesn't exclude the content being copied, per se.

zimeon:
[...]

PR #269, opened 2017-11-22
Clarify ingest-by-reference

awoods:
* Use 'expiration' param on 'message/external-body' for ingest by reference

Resolves: https://github.com/fcrepo/fcrepo-specification/issues/236

### DIFF
--- a/index.html
+++ b/index.html
@@ -539,9 +539,9 @@
           the use of <code>Content-Type: message/external-body</code> values to signal, on POST or PUT, that the Fedora
           server should not consider the request entity to be the LDP-NR's content, but that a
           <code>Content-Type</code> value will signal a name or address at which the content might be retrieved. The
-          <code>url</code> and <code>local-file</code> <code>access-type</code> values motivate this specification, but
-          a Fedora server may support any <code>access-type</code> parameters per the requirements for advertisement
-          and rejection specified here.
+          '<code>url</code>' and '<code>local-file</code>' values of the <code>access-type</code> parameter motivate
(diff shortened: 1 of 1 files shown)


escowles (dismissed):

birkland:
Hm.  Does acceptance of this PR mean that #210 is still in effect, and that the `expiration` parameter needs to be clarified?

To me, it reads like:
> Where N is an integer, vehicles SHOULD respect an `N accidents occur at this intersection per year` sign, if present, by stopping at the intersection.  If safety cannot be accommodated, the driver MUST turn around and seek an alternate route.

A client desiring ingest by reference is likely in the position of (a) picking a value for 'expires' (the specific value almost certainly not meaningful to the client, but needs to be specified anyway.  The client just wants things copied.. now), and (b) hoping the server followed the SHOULD advice and truly copied the content.

At the very least, it would be a good idea to change the wording to say something like _if the server is unable to copy the external content, the request MUST be rejected..._.

escowles:
@birkland That's a good point, it would be better to change "If the <code>expiration</code> parameter cannot be accommodated" to "If the server is unable to copy the external content" to be clear that some other accommodation (like proxying until the expiration time happens) is not acceptable.

awoods:
Wording updated: https://github.com/fcrepo/fcrepo-specification/pull/269/commits/6fb925ea0f903bd955a9fdd9b0edd3c62b61dfd3

In response to: https://github.com/fcrepo/fcrepo-specification/pull/269#issuecomment-346438416

escowles (approved):

zimeon (approved):

dannylamb (approved):

PR #271, opened 2017-11-22
Clarify LDPCv POST

### LINKED ISSUE #228: Accept-Post: */*; p=0.0 clarification
There is the 'Accept-Post' response header that looks like this: `Accept-Post: */*; p=0.0` defined in this section of the spec: https://github.com/fcrepo/fcrepo-specification/blob/master/index.html#L784. I believe it is meant to indicate to the client that on a POST request no body may be submitted. 

Where is what `p` means defined, spec-wise?   Was it made up for this?  If so, you might want to spell out more what `p` means and the ABNF for it. 

Or is it meant to be a `q`? In which case that is not really done per https://tools.ietf.org/html/draft-wilde-accept-post-03#section-3  (actually there are no accept-params for Accept-Post).

zimeon:
Closes #228.

I realize that the suggestion of 415 "Unsupported Media Type" response code to reject POST with a body doesn't make sense in the context of other changes from #228 so I have also changed that to say 4xx plus constrainedBy link.

### DIFF
--- a/index.html
+++ b/index.html
@@ -798,22 +798,20 @@
           <p>
             If an <a>LDPCv</a> supports <code>POST</code>, a <code>POST</code> that does not contain a
             <code>Memento-Datetime</code> header SHOULD be understood to create a new <a>LDPRm</a> contained by the
-            <a>LDPCv</a>, reflecting the state of the <a>LDPRv</a> at the time of the <code>POST</code>.
+            <a>LDPCv</a>, reflecting the state of the <a>LDPRv</a> at the time of the <code>POST</code>. Any request
+            body MUST be ignored.
           </p>
           <p>
-            If an <a>LDPCv</a> supports <code>POST</code>, a <code>POST</code> with a
-            <code>Memento-Datetime</code> header SHOULD be understood to create a new <a>LDPRm</a> contained by the
-            <a>LDPCv</a>, reflecting the state of the request body at the time of the <code>Memento-Datetime</code>
-            request header.
+            If an <a>LDPCv</a> supports <code>POST</code>, a <code>POST</code> with a <code>Memento-Datetime</code>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -798,22 +798,20 @@ <h2>HTTP POST</h2>
...
           </p>
           <p>
-            If an implementation does not support <code>POST</code> with a request body, the <code>Accept-Post</code>
-            header of any response from the <a>LDPCv</a> SHOULD indicate that no request body is accepted via the form
-            <code>Accept-Post: */*; q=0.0</code>, and that implementation MUST respond to any body-containing
-            <code>POST</code> to that <a>LDPCv</a> with a 415 response and a link to an appropriate constraints
-            document (see <a href='https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs'>LDP 4.2.1.6</a>). As per [[!LDP]],
-            any resource created via <code>POST</code>, in this case an <a>LDPRm</a>) SHOULD be advertised in the
-            response's <code>Location</code> header.
+            If an implementation does not support either of <code>POST</code> cases above, it MUST respond to with
```
escowles: "it MUST respond to with" should be "it MUST respond to POST requests with"?

awoods (changes requested):
Pending response to: https://github.com/fcrepo/fcrepo-specification/pull/271#discussion_r152655454

index.html
zimeon: Actually, there is a slightly bigger wording issue in that "either" is ambiguous, I should perhaps say "If an implementation does not support one or both of <code>POST</code> cases above, it MUST respond to such requests with..."

escowles (approved):

dannylamb (approved):

awoods (approved):

PR #285, opened 2017-11-27
define fcrepo:PreferCompactLists IRI for Prefer header values

### LINKED ISSUE #260: Define Prefer header URI for compact collection representation
Turtle and JSON-LD are the only two required serializations in the LDP specification, and both include a compact syntax for ordered collections of statement objects. The specification should propose a Prefer header for requesting this compact form (rather than blank nodes for first and rest).

barmintor:
fixes #260


barmintor:
@dannylamb notes that the language of compaction is ambiguous with regard to JSON-LD

barmintor:
Closing pending post-1.0 discussion of #260

PR #299, opened 2017-11-27
Removing unused use-cases template

escowles:
We created a template, but never collected any use cases.


zimeon (approved):
With regret that we didn't use it

awoods (approved):

PR #318, opened 2018-01-05
Say LDP-RS must be created with LDP-NR on PUT

### LINKED ISSUE #317: Clarify LDPNR creation via PUT
If an LDPNR is created via PUT, is it also required to create an associated LDPRS? This is not [currently specified](https://fedora.info/2017/11/27/spec/#http-put-ldpnr) but I think given it is a requirement for creation via POST then it should also be the case for PUT -- it doesn't make sense for the rules to be different.

zimeon:
Closes #317 

Adds sentence to "Creating resources with PUT" that LDP-RS must be created with LDP-NR, paralleling the POST case.

Also moves the detail in POST section under the LDP-NR sub-section, and renames that from "LDP-NRs" to "Creating LDP-NRs with HTTP POST" because the sub-section isn't about POST to an LDP-NR but about their creation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -473,8 +473,9 @@
             specification.
           </blockquote>
           <p>
-            On creation of an <a>LDP-NR</a>, an implementation MUST create an associated <a>LDP-RS</a> describing
-            that <a>LDP-NR</a> in the same way that it would when <a href="#http-post-ldpnr"></a>.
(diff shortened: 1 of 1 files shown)


escowles (dismissed):

awoods (dismissed):

dannylamb (dismissed):
Looks good.  Thanks!

dannylamb:
@barmintor?

escowles (approved):

awoods (approved):

PR #321, opened 2018-01-17
Avoid suggesting that there should be multiple Link headers

### LINKED ISSUE #312: Clarify many links rather than many link headers
Particularly in [LDPRv GET](https://fcrepo.github.io/fcrepo-specification/#ldprv-get), it seems confusing that we list a set of "Link headers" that are required, when these are best expressed as a set of links in ONE `Link:` header.

This seems misleading/confusing. I note that in LDP the language is often similar and somewhat confusing. Although appealing, the term "link relation" is used in LDP and [RFC5988 sec4](https://tools.ietf.org/html/rfc5988#section-4) is used for the predicate, so I think talking about **links** following [RFC5988 sec3](https://tools.ietf.org/html/rfc5988#section-3) that are expressed in a `Link:` header would be the best option.

zimeon:
Fixes #312

### DIFF
--- a/index.html
+++ b/index.html
@@ -281,11 +281,11 @@
         <section id="ldpnr-ixn-model">
           <h2>LDP-NR creation</h2>
           <p>
-            If, in a successful resource creation request, a <code>Link: rel="type"</code> request header specifies the
-            <a>LDP-NR</a> interaction model (<code>http://www.w3.org/ns/ldp#NonRDFSource</code>, regardless of
-            <code>Content-Type:</code> value), then the server SHOULD handle subsequent requests to the newly created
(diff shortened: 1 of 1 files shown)


awoods:
Travis blues :musical_score: https://travis-ci.org/fcrepo/fcrepo-specification/builds/329738795

awoods (approved):

awoods (dismissed):
Pending Travis fix

dannylamb (approved):
It's a subtle difference, but I think this improves readability, too.

dannylamb:
Assuming you appease our great overlord Travis.

escowles (approved):

barmintor (approved):

PR #336, opened 2018-02-28
LDPCv may disallow PUT/PATCH

### LINKED ISSUE #335: Make PATCH/PUT optional for LDPCv
Currently, we require that LDP-RSs support PATCH and PUT.

* https://fcrepo.github.io/fcrepo-specification/#http-patch:
> Any LDP-RS MUST support PATCH ([LDP] 4.2.7 MAY becomes MUST).

* https://fcrepo.github.io/fcrepo-specification/#http-put-ldprs:
> Any LDP-RS MUST support PUT to update statements that are not server-managed triples (as defined in [LDP] 2). [LDP] 4.2.4.1 and 4.2.4.3 remain in effect.

We explicitly say that LDPCvs don't need to support POST.  

* https://fcrepo.github.io/fcrepo-specification/#ldpcv-post:
> Although an LDPCv is both a TimeMap and an LDPC, it MAY disallow POST requests. 

We do say anything specific about PATCH/PUT for LPDCvs, so I think it is requried.  Do we want to add sections saying that PATCH/PUT support is not required for LDPCvs?

escowles:
Fixes #335

### DIFF
--- a/index.html
+++ b/index.html
@@ -852,6 +852,20 @@
           </blockquote>
         </section>
 
+        <section id="ldpcv-put">
+          <h2>HTTP PUT</h2>
+          <p>
+            Although an <a>LDPCv</a> is both a <a>TimeMap</a> and an <a>LDPC</a>, it MAY disallow PUT requests.
+          </p>
+        </section>
+
+        <section id="ldpcv-patch">
+          <h2>HTTP PATCH</h2>
+          <p>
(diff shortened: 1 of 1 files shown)


zimeon (approved):

awoods (approved):

barmintor (approved):
Thank you!

dannylamb (approved):
:+1:

Issue #308, opened 2017-12-01
Clarify expectations around external body content and Accept-Post response header


sprater:
In the specification, there is this requirement:

> 3.9 External Body Content
> [...]
> Fedora servers MUST advertise support in the Accept-Post response header for each supported access-type parameter value of Content-Type: message/external-body. 

To which HTTP verbs does this requirement apply?  Is there any constraint in terms of type of resource acted upon, or headers sent with the request?  Or is Accept-Post expected in every response header?

escowles:
I would expect the GET, HEAD, and OPTIONS methods for containers to include Accept-Post header. Would updating that section with this wording be clearer?

> GET, HEAD, and OPTIONS responses for LDPCs MUST advertise support in the Accept-Post response header for each supported access-type parameter value of Content-Type: message/external-body.

awoods:
There seems to be some mixed guidance here. The IETF draft indicates `Accept-Post` only on `OPTIONS`:
https://tools.ietf.org/html/draft-wilde-accept-post-00#section-3
> Accept-Post SHOULD appear in the OPTIONS response for any resource that supports the use of the POST method.

However, the language in the LDP Spec broadens this slightly:
https://www.w3.org/TR/ldp/#h-header-accept-post-2

> The Accept-Post HTTP header SHOULD appear in the OPTIONS response for any resource that supports the use of the POST method. The presence of the Accept-Post header in response to any method is an implicit indication that POST is allowed on the resource identified by the Request-URI.

Further, in the examples in the LDP Spec, many (if not all) of the `GET` responses include `Accept-Post`.
https://www.w3.org/TR/ldp/#ldpc-ex-membership-full-liabcont

That said, I would be inclined to require (`MUST`) `Accept-Post` on `OPTIONS`, and suggest (`SHOULD`) on `GET`... noting that `HEAD` is implied with `GET`.

escowles:
👍🏻 to MUST for OPTIONS, and SHOULD for GET/HEAD

sprater:
Sounds good to me.  Shall we create tests on the assumption that the spec will updated accordingly?

awoods:
@sprater : It probably makes sense to move forward based on this discussion. We will have full confidence in the informal consensus here once a PR is committed.

zimeon:
[...]

PR #337, opened 2018-02-28
Detail `Accept-Post` header to be on `OPTIONS` responses.

awoods:
Resolves: https://github.com/fcrepo/fcrepo-specification/issues/308

### DIFF
--- a/index.html
+++ b/index.html
@@ -586,8 +586,9 @@
           <code>URL</code>.
         </p>
         <p id='message-external-body-supported-type'>
-          Fedora servers MUST advertise support in the <code>Accept-Post</code> response header for each supported
-          <code>access-type</code> parameter value of <code>Content-Type: message/external-body</code>.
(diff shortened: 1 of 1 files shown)


awoods:
Note: After reviewing the LDP specification further and considering the complexity of adding `SHOULD` for `Accept-Post` headers on `GET` responses for LDPC's but not LDPCv's, I believe it is cleanest to simply add the point of clarification in the `3.9 External Binary Content` section that `Accept-Post` is to be found in responses to `OPTIONS` requests.

escowles (approved):

zimeon:
Apologies but I have questions back up on #308...

escowles (changes requested):
Given ongoing discussion, I think we should close this PR in favor of pursuing #338

awoods:
Closing in favor of direction discussed in https://github.com/fcrepo/fcrepo-specification/issues/338

PR #340, opened 2018-03-08
Prefer headers request not require

### LINKED ISSUE #339: Prefer headers don't REQUIRE behaviors
Currently, the [Additional values for the Prefer header](https://fcrepo.github.io/fcrepo-specification/#additional-prefer-values) section says:

> * http://www.w3.org/ns/oa#PreferContainedDescriptions: Requires a server to include representations of any contained resources in the response, as defined in [annotation-vocab].
> * http://fedora.info/definitions/fcrepo#PreferInboundReferences: Requires a server to include triples from any LDP-RS housed in that server that feature the requested resource as RDF-object.

Per [RFC7240](https://tools.ietf.org/html/rfc7240#section-2) the `Prefer` header merely requests a behavior:

>  The Prefer request header field is used to indicate that particular server behaviors are preferred by the client but are not required for successful completion of the request.

so I think in both bullets **"Requires"** should be changed to **"Requests"**. I'd also be inclined to change the order of description of these two values, in order to put the SHOULD support one before the MAY support onn.

(Alternatively, if we want to enforce these within Fedora implementations then we would need to point out that departure from the RFC7240 of Prefer. I don't think that was the intention however.)

zimeon:
Closes #339

### DIFF
--- a/index.html
+++ b/index.html
@@ -308,19 +308,19 @@
         <section id="additional-prefer-values">
           <h2>Additional values for the <code>Prefer</code> header</h2>
           <p>
-            In addition to the requirements of [[!LDP]], an implementation MAY support the value
-            <code>http://www.w3.org/ns/oa#PreferContainedDescriptions</code> and SHOULD support the value
-            <code>http://fedora.info/definitions/fcrepo#PreferInboundReferences</code> for the <code>Prefer</code>
-            header when making <code>GET</code> requests on <a>LDPC</a> resources:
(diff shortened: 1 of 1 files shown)


awoods (approved):
This looks good.
It is worth highlighting that the functional difference here is changing `Requires` to `Requests`.

escowles (approved):

dannylamb (approved):
:+1:

barmintor (approved):
please don't have 4 more alternatives in a google doc

PR #341, opened 2018-03-14
message/external-body-> Link header

### LINKED ISSUE #338: Define custom headers for external content requests
After much back and forth discussion has again come around to consideration of custom headers in order to support our community-specific use-cases of creation-by-reference and proxying (replacing use of `Content-type:  message/external-body...`). A proposal based on recent discussions:

> Fedora servers SHOULD support POST or PUT requests to create or update LDP-NRs where the content not the request entity, but is instead to be retrieved from the location specified in the `External-Content` header. A `Prefer` header with value `external-content-handling=MODE` where `MODE` is one of `proxy`, `copy`, or `redirect` is used to specify how the content is to be handled. Thus an example request might have headers:
>
> ``` 
> External-Content: <some-URI>; content-type=<some-content-type>
> Prefer: external-content-handling=<copy|redirect|proxy>
> ```
>
> Fedora servers that support `External-Content` requests MUST advertise the supported handling modes with the `Accept-External-Content-Handling: <proxy|copy|redirect>` header in response to an `OPTIONS` request. 
>
> Fedora servers that support `External-Content` requests MUST treat the `content-type=` parameter of the request in the same way that they would treat the `Content-Type` header of a similar request with content in the request entity.
>
> Fedora servers receiving requests that would create or update an LDP-NR with `External-Content` and an unsupported mode MUST be reject the request with 415 (Unsupported Media Type).

Issues/points:
  * We need to define expectations for `proxy`, `copy`, and `redirect` handling.
  * What is expected or required behavior in response to an `External-Content` with no `Prefer: external-content-handling=MODE` header?
  * Presumably also use the same mechanism for [Replacing Contents from Mementos](https://fcrepo.github.io/fcrepo-specification/#memento-by-reference) which might then include LDP-RS as well as LDP-NR
  * We'd expect support for `http(s)` and `file` URIs for remote and local content respectively. Could a server advertise which is supported? What should response be to a request with an unsupported scheme?
  * This would be an extension of prefer headers. There is no experimental or unregistered space defined by [rfc7240](https://tools.ietf.org/html/rfc7240) or listed in the [IANA registry](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#preferences)

zimeon:
Fixes #338 

Rip out `message/external-body` and replace with link indicating external content, hopefully following the discussion in #338

### DIFF
--- a/index.html
+++ b/index.html
@@ -412,11 +412,8 @@
           </p>
           <blockquote class="informative">
             Non-normative note:
-            Implementations may support <code>Content-Type: message/external-body</code> extensions for request bodies
-            for HTTP <code>POST</code> that would create <a>LDP-NR</a>s. This content-type requires a complete
-            <code>Content-Type</code> header that includes the location of the external body, e.g <code>Content-Type:
-            message/external-body; access-type=URL; URL="http://www.example.com/file"</code>, as defined in
-            [[!RFC2017]]. Requirements for this interaction are detailed in <a href="#external-content"></a>.
+            Implementations may support creation of <a>LDP-NR</a> using <code>POST</code> with external content,
+            see <a href="#external-content"></a>.
           </blockquote>
         </section>
       </section>
@@ -425,9 +422,9 @@
         <h2>HTTP PUT</h2>
         <p>
           Impementations MAY allow the interaction model of an existing resource to be changed by specification of a
-          new LDP type in an HTTP <code>Link: rel="type"</code> header. If supported, requests SHOULD be rejected
-          with a 409 (Conflict) response unless the new LDP type specified is a subtype of the resource's current
-          type.
+          new LDP type in a <code>rel="type"</code> link in the HTTP <code>Link</code> header. If supported, requests
+          SHOULD be rejected with a 409 (Conflict) response unless the new LDP type specified is a subtype of the
+          resource's current type.
         </p>
 
         <section id="http-put-ldprs">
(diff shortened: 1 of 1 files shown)


barmintor (dismissed):

birkland:
What is the difference, from the perspective of the client, of using a `Content-Type` header vs using a `type=` param on the `Link` header.  Do these two alternatives have exactly the same semantics?

zimeon:
@birkland - as currently written the client SHOULD use `type=` because then the client MUST respect it

zimeon:
Have updated to reflect discussion today https://github.com/fcrepo/fcrepo-specification/issues/338#issuecomment-373146067

index.html
```diff
@@ -926,13 +954,18 @@ <h2>Client-Managed Version Creation</h2>
...
+            Using the ingest-by-reference mechanism provided by the <code>external-content-handling=copy</code>
+            preference described in <a href="#external-content"></a>, servers may support replacement of the contents
+            of an <a>LDPRv</a> with that of an <a>LDPRm</a> by providing the <a>LDPRm</a>'s location as the target
+            of a link indicating an external content request. For example, given an <a>LDPRm</a> with URL
+            <code>http://example.org/some/memento</code>, the appropriate headers would be:
           </blockquote>
+          <div class="example">
+<pre>Link: http://example.org/some/memento;
+      rel="http://fedora.info/definitions/fcrepo#ExternalContent"
+Prefer: external-content-handling=copy
```
escowles: This should be updated for the new `handling` parameter syntax.

index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
...
-          Fedora servers receiving requests that would create or update an <a>LDP-NR</a> with
-          <code>Content-Type: message/external-body</code> and an unsupported <code>access-type</code> parameter value
-          MUST respond with 415 (Unsupported Media Type). In the case that a Fedora server does not support external
-          <a>LDP-NR</a> content,
-          all <code>message/external-body</code> messages must be rejected with 415 (Unsupported Media Type).
+        <div class="example">
+<pre>Link: <i>&lt;URL&gt;</i>;
+      rel="http://fedora.info/definitions/fcrepo#ExternalContent";
+      handling="<i>&lt;copy|redirect|proxy&gt;</i>";
+      type="<i>&lt;media-type&gt;</i>"
```
escowles: Would it be clearer to use actual examples here instead of the variables, e.g.,

```
Link: <http://example.org/some/content>; 
      rel="http://fedora.info/definitions/fcrepo#ExternalContent"; 
      handling="proxy"; 
      type="image/tiff"
```

escowles (changes requested):
This looks very good, just a couple of quibbles about the examples.

index.html
zimeon: Indeed!

index.html
zimeon: I agree, will fix

escowles (approved):

PR #342, opened 2018-03-14
Revert "message/external-body-> Link header"

escowles:
Reverts fcrepo/fcrepo-specification#341

PR #343, opened 2018-03-14
message/external-body -> Link header

### LINKED ISSUE #338: Define custom headers for external content requests

zimeon:
Fixes #338 

Replacement for #341 which was accidentally merged then reverted with #342.

Rewrites support for external content to use `Link:` header instead of `Content-Type: message/external-body`. Edits should be up-to-date with discussion up to https://github.com/fcrepo/fcrepo-specification/issues/338#issuecomment-373146067

### DIFF
--- a/index.html
+++ b/index.html
@@ -568,7 +568,7 @@
 
       <section id="external-content">
         <h3>External Binary Content</h3>
-        <blockquote id="message-external-body-variability" class="informative">
+        <blockquote id="external-content-variability" class="informative">
           Non-normative note: Variability among client types and locations may mean that <a>LDP-NR</a> content is
           addressed in ways that are external to the Fedora server but not resolvable by all clients. This specification
           describes the use of a <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> link in a
--- a/index.html
+++ b/index.html
@@ -172,7 +172,7 @@
         <dt><dfn>LDPR</dfn>:</dt>
         <dd>
           A <a href="https://www.w3.org/TR/ldp/#dfn-linked-data-platform-resource">Linked Data Platform Resource</a>
-          as defined in [[!LDP]]. This may be an <a>LDP-RS</a> or a <a>LDP-NR</a>.
+          as defined in [[!LDP]]. This may be an <a>LDP-RS</a> or an <a>LDP-NR</a>.
         </dd>
         <dt><dfn>LDP-RS</dfn>:</dt>
         <dd>
@@ -382,7 +382,7 @@
         <blockquote class="informative">
           Non-normative note:
           Implementations may support creation and update of <a>LDP-NR</a> with external content, and this is
-          advertized with the <code>Allow-External-Content-Handling</code> header, see
+          advertised with the <code>Allow-External-Content-Handling</code> header, see
           <a href="#external-content-options"></a>.
         </blockquote>
       </section>
@@ -618,10 +618,10 @@
         </p>
         <p id='external-content-media-type'>
           Fedora servers MUST use the value of the <code>type</code> attribute in the external content link as
-          the media type of the external content, if provided. Otherwise:
+          the media type of the external content, if provided. Any <code>Content-Type</code> header in the
+          request SHOULD be ignored. If there is no <code>type</code> attribute:
         </p>
         <ul>
-          <li>Servers MAY use the media type specified in a <code>Content-Type</code> header of the request.
           <li>Servers MAY use the media type obtained when accessing the external content via the specified
             scheme (e.g. the <code>Content-Type</code> header for external content accessed via
             <code>http</code>).</li>
@@ -634,7 +634,7 @@
           <a>LDP-NR</a> interaction model in this specification.
         </p>
         <section id='external-content-options'>
-          <h4>Advertizing External Content Support</h4>
+          <h4>Advertising External Content Support</h4>
           <p>
             In addition to the requirements of <a href="#http-options"></a>, servers supporting external content
             requests to an <a>LDPR</a> MUST include an <code>Accept-External-Content-Handling</code> header in the
@@ -954,8 +954,8 @@
          <h2>Replacing Contents from Mementos</h2>
           <blockquote class="informative">
             Non-normative note:
-            Using the ingest-by-reference mechanism provided by the <code>external-content-handling=copy</code>
-            preference described in <a href="#external-content"></a>, servers may support replacement of the contents
+            Using the ingest-by-reference mechanism provided by POST with <code>copy</code> external content handling,
+            as described in <a href="#external-content"></a>, servers may support replacement of the contents
             of an <a>LDPRv</a> with that of an <a>LDPRm</a> by providing the <a>LDPRm</a>'s location as the target
             of a link indicating an external content request. For example, given an <a>LDPRm</a> with URL
             <code>http://example.org/some/memento</code>, the appropriate headers would be:
@@ -963,7 +963,8 @@
           <div class="example">
 <pre>Link: &lt;http://example.org/some/memento&gt;;
       rel="http://fedora.info/definitions/fcrepo#ExternalContent";
-      handling="copy"
+      handling="copy";
+      type="image/tiff"
 </pre>
           </div>
         </section>
--- a/index.html
+++ b/index.html
@@ -382,7 +382,7 @@
         <blockquote class="informative">
           Non-normative note:
           Implementations may support creation and update of <a>LDP-NR</a> with external content, and this is
-          advertised with the <code>Allow-External-Content-Handling</code> header, see
+          advertised with the <code>Accept-External-Content-Handling</code> header, see
           <a href="#external-content-options"></a>.
         </blockquote>
       </section>
@@ -591,12 +591,16 @@
           Fedora servers SHOULD support the creation and update of <a>LDP-NR</a>s with content external to
           the request entity, as indicated by a link with
           <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> and target that is the
-          location of the external content. The <code>handling="<i>&lt;copy|redirect|proxy&gt;</i>"</code>
-          attribute specifies how the server should handle the external content, and the optional
-          <code>type="<i>&lt;media-type&gt;</i>"</code> attribute specifies the media type of the external
-          content. Fedora servers that do not support the creation of <a>LDP-NR</a>s with content external
-          to the request entity MUST reject such requests with a 4xx range status code. Fedora servers MUST
-          reject with a 4xx range status code requests that include multiple
+          location of the external content. The <code>handling</code> attribute specifies how the server
+          should handle the external content, and the optional <code>type</code> attribute specifies the
+          media type of the external content.
+        </p>
+        <p>
+          Fedora servers that do not support the creation of <a>LDP-NR</a>s with content external to the
+          request entity MUST reject such requests with a 4xx range status code, and MUST describe this
+          restriction in a resource indicated by a <code>rel="http://www.w3.org/ns/ldp#constrainedBy"</code>
+          link in the <code>Link</code> response header.
+          Fedora servers MUST reject with a 4xx range status code requests that include multiple
           <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> links.
         </p>
         <p id='external-content-handling'>
@@ -614,7 +618,10 @@
         </ul>
         <p>
           Fedora servers MUST reject with a 4xx range status code requests for which the <code>handling</code>
-          attribute is not present or cannot be respected.
+          attribute is not present or cannot be respected. In the case that the specified <code>handling</code>
+          cannot be respected, the restrictions causing the request to fail MUST be described in a resource indicated
+          by a <code>rel="http://www.w3.org/ns/ldp#constrainedBy"</code> link in the <code>Link</code> response
+          header.
         </p>
         <p id='external-content-media-type'>
           Fedora servers MUST use the value of the <code>type</code> attribute in the external content link as
--- a/index.html
+++ b/index.html
@@ -379,6 +379,12 @@
           Non-normative note: Particular <code>Allow:</code> header responses that clients may use to determine
           implementation support for specific features are described in the appropriate sections of this specification.
         </blockquote>
+        <blockquote class="informative">
+          Non-normative note:
+          Implementations may support creation and update of <a>LDP-NR</a> with external content, and this is
+          advertized with the <code>Allow-External-Content-Handling</code> header, see
+          <a href="#external-content-options"></a>.
+        </blockquote>
       </section>
 
       <section id="http-post">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
...
-          <a href="https://tools.ietf.org/html/rfc2017#section-3">section 3</a>) and <code>local-file</code>
-          ([[RFC2046]] <a href="https://tools.ietf.org/html/rfc2046#section-5.2.3.4">section 5.2.3.4</a>) values of the
-          <code>access-type</code> parameter motivate this specification, but a Fedora server may support any
-          <code>access-type</code> parameters per the requirements for advertisement and rejection specified here.
+          describes the use of a <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> link in a
+          <code>Link</code> header to signal, on <code>POST</code> or <code>PUT</code>, that the Fedora server should
+          not consider the request entity to be the <a>LDP-NR</a>'s content, but instead that the target of the
+          link gives the URI from which the content might be retrieved. The <code>http(s):</code> [[RFC3986]] and
+          <code>file:</code> [[RFC8089]] URI schemes motivate this specification, but a Fedora server may support
+          other URI schemes for addressing external content per the requirements for advertisement and rejection
```
escowles, line 579: I don't believe we specify a mechanism for advertising the URI schemes — should we add an `Accept-External-Content-Scheme` header to `OPTIONS` responses for this?  Or advertise it some other way?

zimeon: True. It seems to me that knowing about support for `file:` as a scheme will never be enough to use it, so advertising support doesn't really matter. However, general support for `http` and/or `https` would be usable without additional information. Off the top of my head I don't see anything better than `Accept-External-Content-Scheme: file,http,https` as you suggest

zimeon: I note that with `message/external-body` there was a way to advertise `file` or `anything-else` but not to distinguish between `http`, `https`, `ftp` etc..

index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
...
+          Fedora servers SHOULD support the creation and update of <a>LDP-NR</a>s with content external to
+          the request entity, as indicated by a link with
+          <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> and target that is the
+          location of the external content. The <code>handling="<i>&lt;copy|redirect|proxy&gt;</i>"</code>
+          attribute specifies how the server should handle the external content, and the optional
+          <code>type="<i>&lt;media-type&gt;</i>"</code> attribute specifies the media type of the external
+          content. Fedora servers that do not support the creation of <a>LDP-NR</a>s with content external
+          to the request entity MUST reject such requests with a 4xx range status code. Fedora servers MUST
+          reject with a 4xx range status code requests that include multiple
+          <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> links.
```
whikloj: I don't see a definition for `handling` here, should there be one?

whikloj: Oh its below, nevermind 😊

index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
...
+        </ul>
+        <p id='external-content-body-response-headers'>
+          A Fedora server receiving requests that would create or update an <a>LDP-NR</a> with content external to
+          the request entity MUST reject request if it cannot guarantee all of the response headers required by the
+          <a>LDP-NR</a> interaction model in this specification.
         </p>
-        <section id='external-content-caveats'>
-          <h4>Referenced RDF Content in Mandatory LDP Serializations</h4>
+        <section id='external-content-options'>
+          <h4>Advertizing External Content Support</h4>
```
bbpennel: Should be "Advertising".  http://grammarist.com/spelling/advertise-vs-advertize/

index.html
```diff
@@ -926,13 +954,18 @@ <h2>Client-Managed Version Creation</h2>
...
             Non-normative note:
-            Using the ingest-by-reference mechanism, one can replace the contents of an <a>LDPRv</a> with that of
-            an <a>LDPRm</a> by providing it's URL as the <code>URL</code> parameter in a <code>Content-Type:
-            message/external-body</code> header.  For example, given an <a>LDPRm</a> with URL <code>
-            http://example.org/some/memento</code>, the full header would be <br/>
-            <code>Content-Type: message/external-body; access-type=URL; expiration=1;<br/>
-            &nbsp;&nbsp;&nbsp;&nbsp;URL="http://example.org/some/memento"</code>
+            Using the ingest-by-reference mechanism provided by the <code>external-content-handling=copy</code>
+            preference described in <a href="#external-content"></a>, servers may support replacement of the contents
+            of an <a>LDPRv</a> with that of an <a>LDPRm</a> by providing the <a>LDPRm</a>'s location as the target
```
bseeger: change "an" to "a"   (2 of these)

index.html
```diff
@@ -926,13 +954,18 @@ <h2>Client-Managed Version Creation</h2>
          <h2>Replacing Contents from Mementos</h2>
           <blockquote class="informative">
             Non-normative note:
-            Using the ingest-by-reference mechanism, one can replace the contents of an <a>LDPRv</a> with that of
-            an <a>LDPRm</a> by providing it's URL as the <code>URL</code> parameter in a <code>Content-Type:
-            message/external-body</code> header.  For example, given an <a>LDPRm</a> with URL <code>
-            http://example.org/some/memento</code>, the full header would be <br/>
-            <code>Content-Type: message/external-body; access-type=URL; expiration=1;<br/>
-            &nbsp;&nbsp;&nbsp;&nbsp;URL="http://example.org/some/memento"</code>
+            Using the ingest-by-reference mechanism provided by the <code>external-content-handling=copy</code>
```
bseeger: Above "copy" only say `requests that the server dereference the external content URI and treat that as if it were the entity body of the request.`   It does not say that it has to be ingested into the Fedora system.  So, "ingest-by-reference" doesn't quite hold here and this ultimately is different functionality.

index.html
escowles: We currently use "an LDP..." 55 of 56 times in the spec.   I don't have a strong feeling about whether we should agree with the acronym or spelled-out form, but we should be consistent.

index.html
```diff
@@ -926,13 +954,18 @@ <h2>Client-Managed Version Creation</h2>
...
+            preference described in <a href="#external-content"></a>, servers may support replacement of the contents
```
birkland: I believe this is no longer a preference, but rather a value of `handling`

index.html
```diff
@@ -926,13 +954,18 @@ <h2>Client-Managed Version Creation</h2>
...
+            of an <a>LDPRv</a> with that of an <a>LDPRm</a> by providing the <a>LDPRm</a>'s location as the target
+            of a link indicating an external content request. For example, given an <a>LDPRm</a> with URL
+            <code>http://example.org/some/memento</code>, the appropriate headers would be:
           </blockquote>
+          <div class="example">
+<pre>Link: &lt;http://example.org/some/memento&gt;;
+      rel="http://fedora.info/definitions/fcrepo#ExternalContent";
+      handling="copy"
```
birkland: This is a good, informative section.  If the server supports the `copy` handling, then the above example is still  _may or may not_ work because of the absence of `type=`?  i.e servers are within their right to reject the request due to the absence of a type, might assign a different (default) mime type to the `LRPRv`,  but might also do the Right Thing implicitly.

index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
...
-          the Fedora server is proxying the content.
+        <p id='external-content-media-type'>
+          Fedora servers MUST use the value of the <code>type</code> attribute in the external content link as
+          the media type of the external content, if provided. Otherwise:
         </p>
-        <p id='external-content-want-digest'>
-          <code>GET</code> and <code>HEAD</code> requests to any external <a>LDP-NR</a> MUST correctly
-          respond to the <code>Want-Digest</code> header defined in [[!RFC3230]].
+        <ul>
+          <li>Servers MAY use the media type specified in a <code>Content-Type</code> header of the request.
```
birkland: So, is there any semantic difference between a request that contains a `Content-Type` but not a `type=` vs a request that omits `Content-Type` but has a `type=`?

If there is no difference, then why is it explicitly spelled out that a server is allowed to ignore a provided `Content-Type` header for external content?   

I can see that having two ways of doing the same thing would be confusing and problematic, especially if both `Content-Type` and `type=` had a value that was different.  Defining a preferred order of precedence is fine, as is picking one over the other.  The spec, though, suggests that in the absence of `type=`, the server can explicitly ignore `Content-Type`, which does not seem intuitive, as it's not explicitly allowed to ignore it in any other context.

index.html
bseeger: I think I'm incorrect on this as well, it sounds like 'an' might the the accepted article, since the starting sound is "el".  So you can ignore my comment.

index.html
zimeon: Indeed, I've got too used to switching all my s to z!

index.html
zimeon: Indeed, should be _"... the `copy` handling described in..."_

index.html
zimeon: Yes, better to include a `type="image/tiff"` -- I suspect that implementations will be written to do the right thing but since we aren't requiring it, it seems better to be explicit

index.html
zimeon: I think we should remove the bullet about using `Content-Type` and instead say that _"Implementations SHOULD ignore any `Content-Type` header on external content requests."_. There seems only to be confusion, but no benefit, from having two ways of specifying

index.html
zimeon: I think `handling="copy"` and POST combined _is_ ingest-by-reference so perhaps what I need to do is mention _"`POST` with `copy` external content handling"_?

index.html
zimeon: Thought it is expanding scope a wee bit, I'll change the one "a LDP-NR" that @escowles mentioned

index.html
zimeon: I don't feel comfortable making a call on this yet so have created #344 in order to allow this PR forward

index.html
birkland: Yes, that would be more clear. The question as to whether there is a semantic difference between `Content-Type` and `type=` is important.  If there _is_, then it might be worthwhile strengthening that to a _MUST ignore any `Content-Type`_.

escowles (dismissed):

index.html
```diff
@@ -571,70 +571,98 @@ <h3>External Binary Content</h3>
         <blockquote id="message-external-body-variability" class="informative">
```
awoods: We should probably change the `id` to something like: `external-content-variability`.

zimeon: Yup, I didn't see that... will update

index.html
```diff
@@ -568,73 +568,101 @@ <h3>Recursive Delete</h3>
...
+          <li>Servers MAY use the media type obtained when accessing the external content via the specified
+            scheme (e.g. the <code>Content-Type</code> header for external content accessed via
+            <code>http</code>).</li>
+          <li>Servers MAY use a default media type.</li>
+          <li>Servers MAY reject the request with a 4xx range status code.</li>
+        </ul>
+        <p id='external-content-body-response-headers'>
+          A Fedora server receiving requests that would create or update an <a>LDP-NR</a> with content external to
+          the request entity MUST reject request if it cannot guarantee all of the response headers required by the
+          <a>LDP-NR</a> interaction model in this specification.
```
awoods, line 641: We should discuss how support for `LDP-NR` headers could be supported in the `redirect` handling case.

zimeon: May I suggest a new issue in order to avoid expando-ing this one ;-) ?

escowles: 👍 to not expanding the scope of this further

escowles (dismissed):

index.html
```diff
@@ -568,73 +568,101 @@ <h3>Recursive Delete</h3>
...
+        </div>
+        <p id='external-content-link'>
+          Fedora servers SHOULD support the creation and update of <a>LDP-NR</a>s with content external to
+          the request entity, as indicated by a link with
+          <code>rel="http://fedora.info/definitions/fcrepo#ExternalContent"</code> and target that is the
+          location of the external content. The <code>handling="<i>&lt;copy|redirect|proxy&gt;</i>"</code>
+          attribute specifies how the server should handle the external content, and the optional
+          <code>type="<i>&lt;media-type&gt;</i>"</code> attribute specifies the media type of the external
+          content. Fedora servers that do not support the creation of <a>LDP-NR</a>s with content external
+          to the request entity MUST reject such requests with a 4xx range status code. Fedora servers MUST
```
zimeon: Eds call: add requirement to link to constraints doc

index.html
```diff
@@ -568,73 +568,101 @@ <h3>Recursive Delete</h3>
...
+            it were the entity body of the request.</li>
+          <li><code>redirect</code> - requests that the server record the location of the external content and
+            handle requests for that content using HTTP redirect responses with the <code>Content-Location</code>
+            header specifying the external content location. See also <a href="#redirect-and-proxy"></a>.</li>
+          <li><code>proxy</code> - requests that the server  record the location of the external content and
+            handle requests for that content by proxying. See also <a href="#redirect-and-proxy"></a>.</li>
+        </ul>
+        <p>
+          Fedora servers MUST reject with a 4xx range status code requests for which the <code>handling</code>
+          attribute is not present or cannot be respected.
```
zimeon: Eds' call: require link to constraints doc if cannot be respected

index.html
```diff
@@ -379,6 +379,12 @@ <h2>HTTP OPTIONS</h2>
           Non-normative note: Particular <code>Allow:</code> header responses that clients may use to determine
           implementation support for specific features are described in the appropriate sections of this specification.
         </blockquote>
+        <blockquote class="informative">
+          Non-normative note:
+          Implementations may support creation and update of <a>LDP-NR</a> with external content, and this is
+          advertised with the <code>Allow-External-Content-Handling</code> header, see
```
zimeon: Typo: should be `Accept-External-Content-Handling`

escowles (approved):

barmintor (approved):

awoods (approved):

awoods:
Note: We should update anchor links in the wiki and in the Test Suite when this PR is merged.
https://wiki.duraspace.org/display/FF/Test+Compatibility+Suite+Verification

dannylamb (approved):
:shipit:

PR #350, opened 2018-04-04
Add Preference-Applied "if and only if all requested preferences were applied"

### LINKED ISSUE #349: Should the `Preference-Applied` response header include parameters?
[Example 20/21 in LDP](https://www.w3.org/TR/ldp/#prefer-examples-direct-minimal-container-only1) and https://tools.ietf.org/html/rfc7240#section-3:

> The syntax of the Preference-Applied header differs from that of the Prefer header in that parameters are not included

suggest that the `Preference-Applied` response header should include only the preference name (i.e. `return=representation`) and not the parameters (e.g. `omit="..."`). This is at odds with the Fedora 4 spec/implementation as shown in the [examples for GET](https://wiki.duraspace.org/display/FEDORA4x/RESTful+HTTP+API+-+Containers#RESTfulHTTPAPI-Containers-GETRetrievethecontentoftheresource).

We have two choices I think:

  1. Follow RFC7240 and LDP, and return only `Preference-Applied: return=representation`
  2. Deviate from RFC7240 and keep Fedora 4 behavior which means we'd need to document that in the API spec.

I think we should do the former as we are trying to follow where possible. (See also https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/issues/50)

zimeon:
Fixes #349

### DIFF
--- a/index.html
+++ b/index.html
@@ -329,7 +329,8 @@
           <p>
             Responses to <code>GET</code> requests that apply a <code>Prefer</code> request header to any <a>LDP-RS</a>
             MUST include the <code>Preference-Applied</code> response header as defined in [[!RFC7240]]
-            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a>.
+            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only if all requested
+            preferences were applied.
           </p>
           <p>
             When a <code>GET</code> request is made to an <a>LDP-RS</a> that describes an associated <a>LDP-NR</a>
--- a/index.html
+++ b/index.html
@@ -329,8 +329,8 @@
           <p>
             Responses to <code>GET</code> requests that apply a <code>Prefer</code> request header to any <a>LDP-RS</a>
             MUST include the <code>Preference-Applied</code> response header as defined in [[!RFC7240]]
-            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only if all requested
-            preferences were applied.
+            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only if the requested
+            preference was applied in full (by applying all specified parameters).
           </p>
           <p>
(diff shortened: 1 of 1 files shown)


awoods (dismissed):

escowles (dismissed):

index.html
```diff
@@ -329,8 +329,8 @@ <h2><a>LDP-RS</a>s</h2>
           <p>
             Responses to <code>GET</code> requests that apply a <code>Prefer</code> request header to any <a>LDP-RS</a>
             MUST include the <code>Preference-Applied</code> response header as defined in [[!RFC7240]]
-            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only if the requested
-            preference was applied in full (by applying all specified parameters).
+            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only all specified parameters
```
dannylamb: if and only *if*

index.html
```diff
@@ -329,7 +329,8 @@ <h2><a>LDP-RS</a>s</h2>
           <p>
             Responses to <code>GET</code> requests that apply a <code>Prefer</code> request header to any <a>LDP-RS</a>
             MUST include the <code>Preference-Applied</code> response header as defined in [[!RFC7240]]
-            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a>.
+            <a href='https://tools.ietf.org/html/rfc7240#section-3'>section 3</a> if and only all specified parameters
```
dannylamb: missing the second if in 'if and only if'

dannylamb: ha, sry didn't mean to say that twice.  fooled by browser caching... plus i'm an idiot.

escowles (approved):

dannylamb (approved):
:shipit:

index.html
zimeon: thanks... you ain't the idiot that missed out the 'if' ;-)

awoods (approved):

barmintor (approved):
thanks for the careful changes @zimeon

PR #355, opened 2018-04-25
Adding non-normative note about versioning and containment

### LINKED ISSUE #315: What does it mean to create a Memento of an LDPC and its associated containment notions?
As currently written, [4. Resource Versioning](https://fcrepo.github.io/fcrepo-specification/#resource-versioning) allows any LDPR to be versioned, and thus we may have original and memento resources for a container. Any contained resource could be contained by the original and/or multiple mementos.

By my reading of [LDP](https://www.w3.org/TR/ldp/), each resource is expected to be contained by zero or one containers, not more. I didn't find an explicit statement saying that it isn't possible to have multiple containment, but language like _["Each resource created in the application or site is created within an instance of one of these container-like entities"](https://www.w3.org/TR/ldp/#ldpc-informative)_  suggests it.

Maybe the reconciliation of these issues is obvious but I'm not getting it (and perhaps other readers wouldn't so we should document it better).  Or maybe these is a real issue here about whether one needs to have ideas of containment/reference counting, whether 'containment' by a memento is somehow 'inactive', or...

escowles:
Fixes #315

### DIFF
--- a/index.html
+++ b/index.html
@@ -927,10 +927,10 @@
             Non-normative note:
             [[!LDP]] is not explicit, but implies that a resource may not be contained by more than one
             <a>LDPC</a>. Implementations may want to take this implication into account when creating an
-            <a>LDPRm</a> of a <a>LDPC</a>, to avoid having any members of the <a>LDPC</a> be contained by both
-            the <a>LDPRv</a> and the <a>LDPRm</a> of the <a>LDPC</a>. Alternatives include having a
-            <a>LDPRm</a>'s containment triple point to the member's <a>LDPRm</a>, or having the
-            <a>LDPRm</a> return triples about the <a>LDPRv</a>, not the <a>LDPRm</a> itself.
+            <a>LDPRm</a> of an <a>LDPC</a>, to avoid having any members of the <a>LDPC</a> be contained by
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -921,6 +921,18 @@ <h2>Replacing Contents from Mementos</h2>
             &nbsp;&nbsp;&nbsp;&nbsp;URL="http://example.org/some/memento"</code>
           </blockquote>
         </section>
+        <section id="versioning-and-containment">
+         <h2>Versioning and Containment</h2>
+          <blockquote class="informative">
+            Non-normative note:
+            [[!LDP]] is not explicit, but implies that a resource may not be contained by more than one
+            <a>LDPC</a>. Implementations may want to take this implication into account when creating an
+            <a>LDPRm</a> of a <a>LDPC</a>, to avoid having any members of the <a>LDPC</a> be contained by both
```
zimeon: a**n** LDPC -- to follow standard form instead of "a LDPC"

zimeon:
Looks great modulo minor comment above

zimeon (approved):

barmintor (approved):
thank you for this, Esme!

awoods (approved):

dannylamb (approved):

Issue #356, opened 2018-05-02
Text Clean-up: Section 3 - Resource Management


awoods:
This issue is to update the text in Section 3 per the comments in the [this Google-Doc](https://docs.google.com/document/d/1n7ZVIRaiYG36Ye0AP4oSTYbDN915box6zVteXciojFc/edit?usp=sharing).

The updates should also include:
1. A `3.1 Introduction`
1. Replacement of "An Implementation" with "Implementations"... as applicable.

PR #364, opened 2018-05-07
Adding Resource Management introduction

escowles:
Refs #356

### DIFF
--- a/index.html
+++ b/index.html
@@ -248,22 +248,21 @@
     <section id="resource-management">
       <h2>Resource Management</h2>
 
-      <section id="resource-management-intro" class="informative">
-        <h2>Introduction</h2>
-        <p>
-          This section defines the core operations for managing content, using [[!LDP]] as a baseline. It
-          extends [[!LDP]] in several ways, including making some optional features mandatory and expanding
-          requirements related to binary content.
-        </p>
-        <p>
-          [[!LDP]] defines a hierarchy of resource types, with a top-level distinction between RDF resources
-          (<a>LDP-RS</a>) and non-RDF resources (<a>LDP-NR</a>). Container (<a>LDPC</a>) is a subtype of RDF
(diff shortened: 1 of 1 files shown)


escowles:
I expect more wordsmithing here than in the main section 3 cleanup, so putting this in its own PR.

awoods (dismissed):

index.html
```diff
@@ -248,6 +248,23 @@ <h2>Conventions used in this document</h2>
...
+      <section id="resource-management-intro" class="informative">
+        <h2>Introduction</h2>
+        <p>
+          This section defines the core operations for managing content, using [[!LDP]] as a baseline. It
+          extends [[!LDP]] in several ways, including making some optional features mandatory and expanding
+          requirements related to binary content.
+        </p>
+        <p>
+          [[!LDP]] defines a hierarchy of resources, with a top-level distinction between RDF resources
+          (<a>LDP-RS</a>) and non-RDF resources (<a>LDP-NR</a>).  RDF resources are further divided into
```
zimeon: I think we should make it clear that not all RDF resources are containers. May also be easier to split into two sentence, something like:

> RDF resources may function as containers (<a>LDPC</a>) of other resources. There are several container subtypes, which are primarily differentiated by how much the container-member relationship can be configured.

escowles: 👍 just pushed an update to split into two sentences.

zimeon (dismissed):

zimeon:
Moving goalposts, no "Introduction" heading per #369

escowles:
Updated to remove Introduction header and use blockquoted style instead.

dannylamb (dismissed):
LGTM

awoods (approved):

dannylamb (approved):
Even better

zimeon (approved):

PR #367, opened 2018-05-09
Add Authorization introduction (section 5)

awoods:
Resolves part of: https://github.com/fcrepo/fcrepo-specification/issues/358

### DIFF
--- a/index.html
+++ b/index.html
@@ -1020,8 +1020,8 @@ Location: http://example.org/document</pre>
       </p>
 
       <blockquote class="informative">
-        This section both adds further refinement by clarifying (MUST, SHOULD, MAY) requirements where [[!SOLIDWEBAC]]
-        offers open guidance, as well as by specifying interactions relative to resources defined by [[!LDP]].
+        This section both clarifies (MUST, SHOULD, MAY) requirements where [[!SOLIDWEBAC]]
+        offers open guidance, as well as specifies interactions relative to resources defined by [[!LDP]].
       </blockquote>
       <section id="solid-ldp-acls">
         <h2>ACLs are LDP RDF Sources</h2>
--- a/index.html
+++ b/index.html
@@ -1015,9 +1015,16 @@ Location: http://example.org/document</pre>
     <section id="resource-authorization">
       <h2>Resource Authorization</h2>
       <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1015,9 +1015,16 @@ <h2>Versioning and Containment</h2>
...
       <p>
-        To configure access control restrictions, implementations MUST follow the recommendations of Web Access
-        Control [[!SOLIDWEBAC]] with the following additional requirements:
+        Implementations MUST follow the recommendations of Web Access Control [[!SOLIDWEBAC]] in configuring access
+        controls.
       </p>
+
+      <blockquote class="informative">
+        The [[!SOLIDWEBAC]] specification provides refinement on the original
+        <a href='https://www.w3.org/wiki/WebAccessControl'>Web Access Control wiki</a>. This section both adds further
```
zimeon: If we want to reference the WAC wiki then it should become a non-normative reference.

But is it really worth referencing? Or just confusing? -- I'd be inclined to omit it

dannylamb: Agreed. If we just mentioned we're regining SOLIDWEBAC, that'd be enough.

index.html
```diff
@@ -1020,10 +1020,8 @@ <h2>Resource Authorization</h2>
       </p>
 
       <blockquote class="informative">
-        The [[!SOLIDWEBAC]] specification provides refinement on the original
-        <a href='https://www.w3.org/wiki/WebAccessControl'>Web Access Control wiki</a>. This section both adds further
-        refinement by clarifying (MUST, SHOULD, MAY) requirements where [[!SOLIDWEBAC]] offers open guidance, as well
-        as by specifying interactions relative to resources defined by [[!LDP]].
+        This section both adds further refinement by clarifying (MUST, SHOULD, MAY) requirements where [[!SOLIDWEBAC]]
```
zimeon: Now the wiki comment is deleted I don't think we can have "further", but perhaps the start could be made simpler anyway. How about:

> This section both clarifies (MUST, SHOULD, MAY) requirements where [[!SOLIDWEBAC]] ...

escowles (approved):

zimeon (approved):

barmintor (approved):

PR #390, opened 2018-07-27
Refine description of constrainedBy

### LINKED ISSUE #389: Change requirement for `constrainedBy` from SHOULD to MUST to match LDP
From https://github.com/fcrepo/fcrepo-specification/issues/380#issuecomment-404267511) ... currently [we have](https://fedora.info/2018/06/25/spec/#constraints-document):

> 3.1.3 Constraints Document
> When implementation choices result in failure to complete a client request, this SHOULD be expressed in a constraints document, by adding a `Link` header with a link relation of `http://www.w3.org/ns/ldp#constrainedBy`, and a target URI identifying a document expressing the constraints, as outlined by ([LDP] 4.2.1.6).

which is weaker than [LDP 4.2.1.6](https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs):

> 4.2.1.6 LDP servers MUST publish any constraints on LDP clients’ ability to create or update LDPRs, by adding a Link header with an appropriate context URI, a link relation of `http://www.w3.org/ns/ldp#constrainedBy`, and a target URI identifying a set of constraints [RFC5988], to all responses to requests that fail due to violation of those constraints. ...

I propose changing SHOULD to MUST in order that these match.

zimeon:
Closes #389

### DIFF
--- a/index.html
+++ b/index.html
@@ -271,10 +271,10 @@
         <section id="constraints-document">
           <h2>Constraints Document</h2>
           <p>
-            When implementation choices result in failure to complete a client request, this SHOULD be expressed in a
-            constraints document, by adding a <code>Link</code> header with a link relation of
-            <code>http://www.w3.org/ns/ldp#constrainedBy</code>, and a target URI identifying a document expressing the
-            constraints, as outlined by ([[LDP]] <a href="https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs">4.2.1.6</a>).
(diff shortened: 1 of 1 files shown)


dannylamb (approved):
This seems like a good change.  We're no longer deviating and it would enforce a good pattern of behaviour. :+1:

awoods (approved):

escowles (approved):

PR #394, opened 2018-08-29
Remove requirement for constrainedBy on successful POST

### LINKED ISSUE #380: Do we intent to require the constraints document link on all responses?
In the [spec sec 3.1.3](https://fcrepo.github.io/fcrepo-specification/#constraints-document), we reiterate [LDP 4.2.1.6](https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs) that a link to the constraints document SHOULD/MUST be added to responses in the case that the server constrains cause a failure (oddly we say SHOULD when I think it is MUST per LDP). LDP says the header MAY be added to other responses, we are silent on that.

However, in [spec sec 3.5 POST](https://fcrepo.github.io/fcrepo-specification/#http-post) we say:

>  The default interaction model that will be assigned when one is not specified must be recorded in a constraints document.

without any mention of whether one is expected to be able to find that document without making a failing request. One interpretation of this is that the `constrainedBy` link must be present on successful response as well (for POST at least) and this is the interpretation taken by the [test suite in test 3.5-B](https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/c30f47aa55c89ed1336cf4944aff104ce5e6bfa5/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPost.java#L79-L93). It is certainly legal for an implementation to do that but the spec seems unclear at present. I think there are two options:

  * Change  [spec sec 3.1.3](https://fcrepo.github.io/fcrepo-specification/#constraints-document) to say that Fedora implementations should return a link to the constraints doc in all responses
  * Clarify  [spec sec 3.5](https://fcrepo.github.io/fcrepo-specification/#http-post) to say that Fedora implementations are required to return a link to the constraints doc only on failure responses

zimeon:
Closes #380

### DIFF
--- a/index.html
+++ b/index.html
@@ -386,9 +386,7 @@
         <p>
           Any <a>LDPC</a> except <a href="#ldpcv-post">Version Containers (LDPCv)</a> MUST support <code>POST</code>
           ([[!LDP]] <a href="https://www.w3.org/TR/ldp/#ldpr-HTTP_POST">4.2.3</a> /
-          <a href="https://www.w3.org/TR/ldp/#ldpc-HTTP_POST">5.2.3</a>). The default interaction model that will be
-          assigned when one is not specified must be recorded in a <a href="#constraints-document">constraints
-          document</a>.
+          <a href="https://www.w3.org/TR/ldp/#ldpc-HTTP_POST">5.2.3</a>).
         </p>
         <section id="http-post-ldpnr">
           <h2>Creating LDP-NRs with HTTP POST</h2>


escowles (approved):

awoods (changes requested):
Should the `Non-normative` text in [section 4.3.3.2](https://fedora.info/2018/06/25/spec/#ldpcv-post-not-allowed) also be updated?

awoods (approved):

dannylamb (approved):

Issue #393, opened 2018-08-29
memento:OriginalResource Interaction Model Qs


kefo:
The spec introduces a rel="type" Link header for a (new) versioning interaction model: http://mementoweb.org/ns#OriginalResource.  It is mentioned in a single sentence near the beginning of the versioning section:  https://fcrepo.github.io/fcrepo-specification/#resource-versioning

It might be worth considering adding some kind of cross reference to this new interaction model to sections 3.5 (Resource Management: PUT) and 3.6 (Resource Management: POST)

Also, section [3.7.2 (Resource Management, PATCH, Implementation models)](https://fcrepo.github.io/fcrepo-specification/#http-patch-ixn-models) notes that "The server MUST disallow a PATCH request that would change the LDP interaction model of a resource to a type that is not a subtype of the current resource type."  (This is also stated under [3.6 - Resource Management, PUT] (https://fcrepo.github.io/fcrepo-specification/#http-put) - but for some reason does not get its own sub-heading.)  Is "http://mementoweb.org/ns#OriginalResource" a subtype of ldp:BasicContainer, ldp:DirectContainer, ldp:IndirectContainer?  If so, how is this known?  If not, how do we know?

I raise this because, if an implementation does not automatically make all LDPRs versionable (LDPRvs), the client really needs to know to create a versionable resource at the time the resource is created by invoking the memento:OriginalResource interaction model *OR* - I feel - some clarity is needed about whether a subsequent PATCH or PUT request to an existing LDPR can convert it into a LDPRv.

It is also possible that I am overthinking this.

awoods:
My understanding is that `memento:OriginalResource` has no ontological relationship with any LDP types;   and therefore, can be added or removed via `PUT` and the inclusion or exclusion of the appropriate `Link` header without conflicting with existing wording in the specification.

kefo:
My first thought is, I agree that `memento:OriginalResource` has no ontological relationship to the other LDP types, but it's the spec that is potentially creating one.

My second thought is: Does the interaction model for OriginalResource *alter* the BasicContainer implementation model, for example, either by adding functionality or, certainly, removing any?

Related:  https://github.com/fcrepo/fcrepo-specification/issues/229

PR #397, opened 2018-09-14
Add MAY statement for support PUT to convert LDPR into LDPRv

awoods:
Resolves: https://github.com/fcrepo/fcrepo-specification/issues/393


zimeon (approved):

escowles (approved):

acoburn:
What happens if an implementation doesn't support this, and the `PUT` request simply removes the content of the resource? For example, in an implementation that supports server-managed versioning. It seems that this interaction pattern could have some really bad consequences unless a client knows _a priori_ how the server will respond.

zimeon:
@acoburn the PR doesn't change other semantics of PUT. IMO an empty PUT would delete the content of the current resource whether or not the versioning status is changed

acoburn:
Oh, I see. In this case, the `PUT` would contain the resource content; it would just be a new header that is added to the request. Thanks @zimeon!

dannylamb (approved):

PR #399, opened 2018-09-19
Adding section on state tokens

### LINKED ISSUE #396: Specify If-Unmodified-Since-Subsecond header?
The vagaries of RDF serialization prevent us from using ETags for conditional PUT/PATCH requests, leaving the `If-Unmodified-Since` header as the main alternative.  However, the specification for `If-Unmodified-Since` requires a date format that does not allow fractional seconds.  Implementers have found 1-second granularity insufficient — e.g., the Valkyrie optimistic locking implementation for Fedora 4 resorted to supplementing `If-Unmodified-Since` with a finer-grained in-memory lock, to avoid having to insert pauses in their test suite.

Should we specify an alternative header that allows finer granularity, called `If-Unmodified-Since-Subsecond`?

escowles:
Fixes #396

### DIFF
--- a/index.html
+++ b/index.html
@@ -694,7 +694,7 @@ Location: http://example.org/document</pre>
           in resource state without having to compare the entire entity body, and to perform conditional updates.
           Like ETags, the principle behind state tokens is that implementations have the best information on
           resource state, and can choose the most accurate and efficient validation mechanism. State tokens differ
-          from ETags in that there is no expectation the state token must change if the resource is not
+          from ETags in that there is no expectation that the state token must change if the resource is not
           byte-for-byte identical, when the underlying state of the resource has not changed. For example, changes
           in the order of RDF triples, blank node identifiers, and other serialization variations would not require
           the X-State-Token value to change.
@@ -720,14 +720,13 @@ Location: http://example.org/document</pre>
             the client's value.
           </p>
           <p>
-            An HTTP <code>PATCH</code> or <code>PUT</code> request that includes an <code>X-If-State-Token</code>
-            header MUST be rejected with a 400 (Bad Request) response if the implementation does not support state
-            tokens.
+            If an implementation does not support state tokens, it MAY ignore any <code>X-If-State-Token</code> header
+            in HTTP <code>PATCH</code> or <code>PUT</code> requests.
           </p>
           <p>
             An HTTP <code>PATCH</code> or <code>PUT</code> request that includes an <code>X-If-State-Token</code>
-            header MUST be rejected with a 409 (Conflict) response if the implementation supports state tokens, but
-            the client-supplied value does not match the resource's current state token.
+            header MUST be rejected with a 412 (Precondition Failed) response if the implementation supports state
+            tokens, but the client-supplied value does not match the resource's current state token.
           </p>
         </section>
       </section>
--- a/index.html
+++ b/index.html
@@ -686,6 +686,51 @@ Location: http://example.org/document</pre>
           </p>
         </section>
       </section>
+
+      <section id="state-tokens">
+        <h3>State Tokens</h3>
+        <blockquote class="informative">
+          Non-normative note: State tokens are similar to ETags, providing a mechanism for clients to track changes
+          in resource state without having to compare the entire entity body, and to perform conditional updates.
+          Like ETags, the principle behind state tokens is that implementations have the best information on
+          resource state, and can choose the most accurate and efficient validation mechanism. State tokens differ
+          from ETags in that there is no expectation the state token must change if the resource is not
+          byte-for-byte identical, when the underlying state of the resource has not changed. For example, changes
+          in the order of RDF triples, blank node identifiers, and other serialization variations would not require
+          the X-State-Token value to change.
+        </blockquote>
+        <blockquote class="informative">
+          Non-normative note: A state token can be more reliable for validation than an ETag when an implementation
+          cannot provide strong ETags for resources. A state token can be more reliable for validation than a
+          modification date in situations where the one-second resolution of HTTP date values is not sufficient.
+        </blockquote>
+        <section id="x-state-token">
+          <h3>X-State-Token</h3>
+          <p>
+            Implementations MAY include the <code>X-State-Token</code> header field in <code>GET</code> and
+            <code>HEAD</code> responses to provide a token representing the current state of resource. If provided,
+            this value MUST change whenever the underlying state of the resource has changed.
+          </p>
+        </section>
+        <section id="x-if-state-token">
+          <h3>X-If-State-Token</h3>
+          <p>
+            A client MAY include the <code>X-If-State-Token</code> header field in a <code>PATCH</code> or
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
       </section>
+
+      <section id="state-tokens">
+        <h3>State Tokens</h3>
+        <blockquote class="informative">
+          Non-normative note: State tokens are similar to ETags, providing a mechanism for clients to track changes
+          in resource state without having to compare the entire entity body, and to perform conditional updates.
+          Like ETags, the principle behind state tokens is that implementations have the best information on
+          resource state, and can choose the most accurate and efficient validation mechanism. State tokens differ
+          from ETags in that there is no expectation the state token must change if the resource is not
```
awoods: Grammar: "...there is no expectation that [a? the?] state token must change..."

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+            the client's value.
+          </p>
+          <p>
+            An HTTP <code>PATCH</code> or <code>PUT</code> request that includes an <code>X-If-State-Token</code>
+            header MUST be rejected with a 400 (Bad Request) response if the implementation does not support state
+            tokens.
+          </p>
+          <p>
+            An HTTP <code>PATCH</code> or <code>PUT</code> request that includes an <code>X-If-State-Token</code>
+            header MUST be rejected with a 409 (Conflict) response if the implementation supports state tokens, but
```
awoods: `412 (Precondition Failed)` is more common in this case, no?

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+        <section id="x-if-state-token">
+          <h3>X-If-State-Token</h3>
+          <p>
+            A client MAY include the <code>X-If-State-Token</code> header field in a <code>PATCH</code> or
+            <code>PUT</code> request to make the request conditional on the resource's current state token matching
+            the client's value.
+          </p>
+          <p>
+            An HTTP <code>PATCH</code> or <code>PUT</code> request that includes an <code>X-If-State-Token</code>
+            header MUST be rejected with a 400 (Bad Request) response if the implementation does not support state
```
awoods: Lack of support for `X-State-Token` is implicit by the implementation not including the header in responses. That being the case, I would suggest:
```
...MAY be ignored if the implementation does not support state tokens.
```

awoods (changes requested):

index.html
zimeon: I agree, would match [If-Match](https://tools.ietf.org/html/rfc7232#section-3.1) use of 412

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
           </p>
         </section>
       </section>
+
+      <section id="state-tokens">
+        <h3>State Tokens</h3>
+        <blockquote class="informative">
+          Non-normative note: State tokens are similar to ETags, providing a mechanism for clients to track changes
```
zimeon: Perhaps change "track changes in resource state" to "detect a change in resource state" to avoid overstating what one gets from a token

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+          in resource state without having to compare the entire entity body, and to perform conditional updates.
+          Like ETags, the principle behind state tokens is that implementations have the best information on
+          resource state, and can choose the most accurate and efficient validation mechanism. State tokens differ
```
zimeon: I find the use of "validation" jarring as it suggests to me something more than what a state token does. Perhaps "validation mechanism" -> "change detection mechanism" (although really we want to talk about detecting NO change)

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+          from ETags in that there is no expectation the state token must change if the resource is not
+          byte-for-byte identical, when the underlying state of the resource has not changed. For example, changes
+          in the order of RDF triples, blank node identifiers, and other serialization variations would not require
+          the X-State-Token value to change.
+        </blockquote>
+        <blockquote class="informative">
+          Non-normative note: A state token can be more reliable for validation than an ETag when an implementation
```
zimeon: "validation" again... perhaps "A state token can provide reliable indication of an unchanged resource in cases where implementations could not provide strong ETags."

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+          cannot provide strong ETags for resources. A state token can be more reliable for validation than a
```
zimeon: "validation" again... perhaps "A state token can provide a more reliable indication of an unchanged resource than modification date in situations where the one-second resolution of HTTP date values is not sufficient."

zimeon (changes requested):

awoods (dismissed):

escowles:
I've updated the PR to remove the "validation" language and instead use "detect changes in state" etc.

dannylamb (dismissed):
:+1: This is a great feature to provide conditional updates for LDP-RSs

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+
+      <section id="state-tokens">
+        <h3>State Tokens</h3>
+        <blockquote class="informative">
+          Non-normative note: State tokens are similar to ETags, providing a mechanism for clients to detect changes
+          in resource state without having to compare the entire entity body, and to perform conditional updates.
+          Like ETags, the principle behind state tokens is that implementations have the best information on
+          resource state, and can choose the most accurate and efficient mechanism for clients to detect changes.
+          State tokens differ from ETags in that there is no expectation that the state token must change if
+          the resource is not byte-for-byte identical, when the underlying state of the resource has not changed.
```
zimeon: Sorry for not noticing this earlier... but ETags are for a _representation_, or _entity body_ as we have been saying, rather than resources. I think fix is easy (2 places). Here: "...**the entity body** is not byte-for-byte identical"

index.html
```diff
@@ -686,6 +686,51 @@ <h4>Redirected and Proxied External Content</h4>
...
+          For example, changes in the order of RDF triples, blank node identifiers, and other serialization
+          variations would not require the <code>X-State-Token</code> value to change.
+        </blockquote>
+        <blockquote class="informative">
+          Non-normative note: State tokens can be a more reliable mechanism for detecting changes in resource state
+          than ETags when an implementation cannot provide strong ETags for resources. State tokens can be a more
```
zimeon: suggest deleting "for resources" in _"... cannot provide strong ETags ~for resources~."_ (seems easier than mentioning entity bodies again)

zimeon (changes requested):

index.html
escowles: Good catch @zimeon — I'll update the PR.

dannylamb (approved):

zimeon (approved):

awoods (approved):

PR #401, opened 2018-09-19
Add rules for setting `ldp:membershipResource`, `ldp:hasMemberRelation`, `ldp:insertedContentRelation`

### LINKED ISSUE #387: Rules for setting `ldp:membershipResource`,  `ldp:hasMemberRelation`, `ldp:insertedContentRelation`
By my reading, LDP is silent on how container properties are actually set and under what conditions they may be changed. I think we probably all have a working assumption that these values might be set by the RDF supplied on container creation (and I notice this is embodied in [tests such as 3.1.1-C](https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/3c4292c0f8c0ef94f2f0f271e9d31fd123561f50/src/main/java/org/fcrepo/spec/testsuite/crud/Container.java#L133-L137)). Do we need to spell that out? And what/when can changes be made?

zimeon:
Closes #387

### DIFF
--- a/index.html
+++ b/index.html
@@ -259,6 +259,68 @@
             predicate, and thus triples with this predicate do not necessarily imply containment.
           </blockquote>
         </section>
+        <section id="ldpdc">
+          <h2>LDP Direct Containers</h2>
+          <p>
+            Implementations that support the creation and management of LDP Direct Containers ([[!LDP]]
+            <a href="https://www.w3.org/TR/ldp/#ldpdc">5.4</a>)
+            use the <code>ldp:membershipResource</code> property to express the membership constant URI (per
+            [[!LDP]] <a href="https://www.w3.org/TR/ldp/ldpdc-containres">5.4.1.3</a>), and use either the
+            <code>ldp:hasMemberRelation</code> or <code>ldp:isMemberOfRelation</code> property to express
+            the member relation (per [[!LDP]] <a href="https://www.w3.org/TR/ldp/#ldpdc-containtriples">5.4.1.4</a>).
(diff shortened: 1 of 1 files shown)


awoods (dismissed):

index.html
```diff
@@ -259,6 +259,68 @@ <h2>LDP Containers (LDPC)</h2>
             predicate, and thus triples with this predicate do not necessarily imply containment.
           </blockquote>
         </section>
+        <section id="ldpdc">
+          <h2>LDP Direct Containers</h2>
+          <p>
+            Implementations that support the creation and management of LDP Direct Containers ([[!LDP]]
+            <a href="https://www.w3.org/TR/ldp/#ldpdc">5.4</a>)
+            use the <code>ldp:membershipResource</code> property to express the membership constant URI (per
+            [[!LDP]] <a href="https://www.w3.org/TR/ldp/ldpdc-containres">5.4.1.3</a>), and use either the
```
dannylamb: Link does not resolve.  Should be https://www.w3.org/TR/ldp/#ldpc ?

escowles (dismissed):

dannylamb (changes requested):

index.html
zimeon: Oh, yes, missing #... should be https://www.w3.org/TR/ldp/#ldpdc-containres . Thanks, will fix

dannylamb (approved):

escowles (approved):

awoods (approved):

barmintor (approved):

PR #412, opened 2021-01-21
Add additional Prefer header: ServerManaged

awoods:


zimeon:
I don't see any problem adding the `http://fedora.info/definitions/fcrepo#ServerManaged` URI in order to talk about the server managed triples if there are use cases for that. If I'm correct, these would be included by default so only `omit` is useful.

However, I think changing the intro para to mention both `include` and `omit` (as opposed to just `include` which is useful with the other two values) opens a whole can of worms about possible implementation. LDP is extremely vague about exactly how `include` and `omit` hints should be processed or interact ( https://www.w3.org/TR/ldp/#prefer-parameters).

With the existing URI values we try to clarify this by mentioning use only with `include`.

Should the new bullet for `http://fedora.info/definitions/fcrepo#ServerManaged` be changed to mention use only with `omit`? Alternatively, the default behavior -- and thus the useful use of the parameters -- could be made explicit in all cases. 

I think at this stage I'd be happiest with minimal change which would be `s/response include or omit triples/response omit triples/` in the PR

zimeon (changes requested):
See https://github.com/fcrepo/fcrepo-specification/pull/412#issuecomment-764868570

escowles:
I agree with @zimeon — changing the text to limit to omitting server-managed triples would be good, to avoid confusion about which URIs can be used with omit vs. include options.

awoods:
> If I'm correct, these would be included by default so only `omit` is useful.

correct

The PR has been updated to only mention `omit` for the new header value. If it would be helpful to modify the phrasing of the intro text to more clearly indicate that the first two header values only relate to `include` parameters and the third header value to `omit`, I can attempt to draft something.

escowles (approved):
👍🏻 This seems clear to me — each URI uses "include" or "omit" to indicate which operation is appropriate.

zimeon (approved):

dannylamb (approved):
:+1: to these changes

awoods:
[Issue Resolution Process](https://github.com/fcrepo/fcrepo-specification/wiki/Issue-Resolution-Process)
> 8. A pull request may be merged when all editors have responded with 👍/+1, or after 72 hours have passed and a majority of editors have responded with 👍/+1, and no editors have responded with 👎/-1.

PR #413, opened 2021-04-27
Release 1.0

zimeon:
Prep for 1.0 release having sat as candidate since 2018!

I have chopped a whole bunch out of reSpec to avoid lots of W3C stuff getting inappropriately added.

PR #414, opened 2021-04-30
Fix auth/acl ns scheme

csarven:


zimeon:
Hmm... if I dereference https://www.w3.org/ns/auth/acl I get a document that includes...
```
    @prefix acl: <http://www.w3.org/ns/auth/acl#> .
```

csarven:
w3.org is making https available all across (AFAIK) and supports upgrading insecure requests. So, both http and https URLs of the same document is available. `http:` is the canonical scheme for the auth/acl ns.

zimeon:
Yes, indeed! Sorry, I misread the PR as suggesting changing the ns to https: when in fact you are correcting it to http:. Completely agree

zimeon (approved):

awoods (approved):
Let's get this in before publishing the 1.0 spec.

awoods:
I have published the 1.0 release with this change: https://fedora.info/spec/

We should merge this soon... although, in case there is any concern with including this PR, I will hold off on a community announcement until the merge has happened. I think the `https` to `http` update is non-controversial.

csarven:
Please consider https://github.com/fcrepo/fcrepo-specification/pull/413#issuecomment-830141870 - if/when updates to references and other changes can be introduced.

dannylamb (approved):
sorry i missed this one

